PR #30, opened 2020-11-02
fixed duplication bug

bmorris-revacomm:
Created variable "currentForm" on Form781Controller - when hitting the save button, it will now check if the variable is nil or not. If nil, it will create new object - if not nil, it will update object


petermisik (approved):
Changes look good.

PR #37, opened 2020-11-05
Created alert for invalid time

petermisik:


petermisik:
Build doesn't work.  Pushed too quickly.

PR #41, opened 2020-11-05
fixed tableview bug

bmorris-revacomm:
refactored create function in form781controller - renamed and moved loadFlights and loadCrewMembers into that function


petermisik (approved):
Everything looks great.

PR #55, opened 2020-11-20
Jira PR-73 A little better checking on date entry.

Pete-Hoch:
Favor day-month-year, but accept month-day-year.

We should probably add an alert that we reverted the date back to the old valid date.

I also changed some of the functions in Helper tp static so that we don't have to instantiate a copy of Helper in order to call them. But I dod not go through and change everything.


petermisik (approved):
I like the extension and changes.

bmorris-revacomm (approved):
Looks great! I was actually going to ask about making all the Helper methods static, but you beat me to it! I assume it's probably better than instantiating Helper each time - is that right?

PR #57, opened 2020-11-20
Just want to get all of the changes on dev over to master (since that seems to be where the action is).

Pete-Hoch:
Just a merge review do get origin/dev merged into origin/master.


petermisik (approved):
Looks solid to me.

PR #59, opened 2020-11-24
Fix/pete m fix unittest

petermisik:
Fixed the unit test checking for Logging.json.  If the file doesn't exist (ie new installation) it'll create a file using generic data using Form781Controller.create function.  Then it'll test again looking for the file.  The XCTAssert is true (file found).


Pete-Hoch (approved):
We have a clean uint test suite!

bmorris-revacomm (approved):
bueno!

PR #72, opened 2020-12-04
Fix/pr72

petermisik:


Pete-Hoch (commented):
As far as getting the print button active, this is good. However;

1) Can't select to print two sided.
2) Can't select a subset of the pages to print.
3) It prints VERY slowly.
4) The actual printed page looks like it's set to a 25% zoom.

petermisik:
Print system has been updated and is faster.  Two sided printing isn't working with my printer, but it could be something with my printer.  

The subset of pages has been removed by Apple since iOS 10.  That's the warning I kept getting when I tried to specify that function in the code.  

Switch from .photo to .general speeds up the print considerably.

Pete-Hoch (approved):
Looks good. Printed in under 2 minutes.

We may want to switch to short edge binding for duplex, but that's minor and we can deal with it later.

PR #73, opened 2020-12-04
Make the forms list private to the forms controller.

Pete-Hoch:
In looking at persistence, I noticed that the 'knowledge' about how we store the 781 forms (and crew data, etc.) was scattered all over the place. This is a small step to hide how we store the data and make formal accessors. We can still do more in this area, but this is a start.


petermisik (approved):
These look great.  Thank you for locking this down.

PR #74, opened 2020-12-07
Fix/added scroll view for keyboard

bmorris-revacomm:
- added scroll view so keyboard doesn't block textfields (there are a couple little UI bugs that I am looking at fixing - they don't affect the functionality at all)
- return key on keyboard now dismisses keyboard
- created openPopUp function to clean up code
- deleted Aircrew storyboard


Logging/Controllers/View Controllers/Form 781/ContainerViews/MissionDataViewController.swift
```diff
@@ -81,8 +81,31 @@ class MissionDataViewController: UIViewController {
         flightSeqTableView.delegate = self
         flightSeqTableView.dataSource = self
         dateTextField.delegate = self
+<<<<<<< HEAD
```
petermisik, line 84: You have a conflict header here you need to resolve.

petermisik (changes requested):
If you resolve the conflict, I'm good with everything else.

Logging/Controllers/View Controllers/Form 781/ContainerViews/MissionDataViewController.swift
bmorris-revacomm: Yes, I saw that and just fixed it!

petermisik (approved):
Everything looks good.

PR #77, opened 2020-12-09
Added conditional to make sure there's atleast 1 form.

petermisik:
Added a second conditional to by pass creating the form if there are no flights allowing them to print a form with just the header information.


cbrech:
Are there unit tests that cover this?

cbrech (approved):

PR #80, opened 2020-12-10
Changed to landscape only

cbrech:
Based on feedback session with Christian, Rachel, and Kelsey


Pete-Hoch (approved):
Simple change, looks good.

PR #81, opened 2020-12-11
Feature/core data migration

Pete-Hoch:
This is the first pass at Core Data persistence. 

I have not worked on the unit tests because I wanted to make sure we would go this route.

Also, this code does not migrate the old persistence forward. It's not too much work, but I didn't want to spend the time if we decide to just loose everyones old data.


cbrech:
I am very wary of Core Data, upgrading the model has to be done just right or the user looses all their data. Has this gotten better/easier in the past two years?

cbrech:
Per stand up 11 dec, closing for now. Will re-open after 0.1.0

PR #87, opened 2020-12-12
Added PB Logo & layout constraints for top bar

cbrech:


Logging/Resources/Assets.xcassets/PB_Logo.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
...
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
```
Pete-Hoch, line 14: OIC, we are just using one image for all resolutions. I think there is a way to just say All in the xcassets for that.

Pete-Hoch (approved):
Looks good.

Logging/Resources/Assets.xcassets/PB_Logo.imageset/Contents.json
cbrech: Great point. I'll try and remember to clean that up

PR #88, opened 2020-12-14
Fix/dim view bug fixed

bmorris-revacomm:
- fixed dim view UI bug
- fixed chevron directions
- menu now starts open


Pete-Hoch (approved):
Looks good.

PR #89, opened 2020-12-14
Fix/pr46 pr47 email pdf

petermisik:
Have share menu incorporated.


Pete-Hoch (approved):
I don't seem to have mail on my simulator, but I did copy and paste into messages.

All I got was: 

Form 781

Current 781

So if that's what I am suppose to get, it worked well.

PR #90, opened 2020-12-14
Project file clean up.

Pete-Hoch:
I noticed that we were including Logger source in the test targets. We do not need to do this, we import Logger as a @testable in the test target code. So this cleans that up.


petermisik (approved):
Thank you for the clean up.

PR #91, opened 2020-12-14
Fix/main view resizes

bmorris-revacomm:
- main view resizes instead of being hidden by menu (some UI issues on smaller ipad - squished labels)


Pete-Hoch (approved):
Looks good.

PR #92, opened 2020-12-14
Prevent crash when less than 4 characters entered for time

Pete-Hoch:
In fact, only accept 4 character time and prompt the user if they enter less or more.


petermisik (approved):
I like the changes.  I should have caught this in the beginning.

PR #94, opened 2020-12-14
reduce to iOS 13.5

cbrech:
Reducing to iOS 13.5 to support now secure testing :(

### DIFF
--- a/Logging.xcodeproj/project.pbxproj
+++ b/Logging.xcodeproj/project.pbxproj
@@ -882,6 +882,7 @@
 				CURRENT_PROJECT_VERSION = 32;
 				DEVELOPMENT_TEAM = RQ9T47YT3S;
 				INFOPLIST_FILE = Logging/Resources/Info.plist;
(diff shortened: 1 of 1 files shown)


Logging.xcodeproj/project.pbxproj
```diff
@@ -882,6 +882,7 @@
 				CURRENT_PROJECT_VERSION = 32;
 				DEVELOPMENT_TEAM = RQ9T47YT3S;
 				INFOPLIST_FILE = Logging/Resources/Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.5;
```
Pete-Hoch, line 885: Just curious, if we need to support iOS 13, why don't we just support 13.0? Apple will rarely add an SDK change on the .X releases. They are typically for bug fixes.

cbrech: 13.5 is their rigs so I set to that, but Ill lower to 13

PR #98, opened 2020-12-16
PR-203: Add notification for data change.

Pete-Hoch:
This is the start of separating UI updates from data modification.

Added a single notification for when the current form changes. This can be used to update the views after we change which form is current as well as it's being used here, to notify all view controllers (who are listening) that the current form's data has changed.

### DIFF
--- a/Logging/Controllers/Model Controllers/Form 781/Form781Controller.swift
+++ b/Logging/Controllers/Model Controllers/Form 781/Form781Controller.swift
@@ -14,6 +14,9 @@ class Form781Controller {
     
     static let shared = Form781Controller()
     
+    // MARK: - Notifications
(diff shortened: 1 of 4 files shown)


cbrech (commented):
Do we need to remove the listeners when the view is closed? Or does Swift handle that automatically now?

Pete-Hoch:
> Do we need to remove the listeners when the view is closed? Or does Swift handle that automatically now?

Since it's the view controller that is listening, we do not have to remove it. This is because in our case, the view controller is never released. However, if it were, I believe the Notification Center has a weak reference to the 'self' and so the listening is effectively removed.

Alternately, we can listen and stop listening when the view is displayed and hidden respectively, but that seems more fragile to me.

cbrech (approved):

PR #99, opened 2020-12-16
Fix/refactored aircrew list pop up

bmorris-revacomm:
(replacing previous PR)
- fixed closed menu constraints
- refactored aircrew list pop up (will do other pop ups later)

Main changes from refactor:
- whole screen is dimmed out and disabled
- pop up is it's own view controller now which will allow bigger and better resizing, and allows pop up to be centered to the screen instead of previous container view (lots of protocols and delegates)


petermisik (approved):
Looks like a solid fix.

PR #102, opened 2020-12-16
Fix/refactored mission data pop up and flight seq pop up

bmorris-revacomm:
(for some reason, it looks like my previous PR commits are in here - not sure why)

- refactored mission data pop up
- refactored flight seq pop up

@Pete-Hoch let me know if these changes affect the currentFormChanged func you added. I also moved the location of your reloadCurrentFormViews because of the refactor - let me know how that looks


Pete-Hoch (approved):
Works really nicely.

PR #103, opened 2020-12-16
Set share menu to point to SEND button.

Pete-Hoch:
Once I saw that this was a view sizing issue, I googled this: "UIActivityViewController wrong size" and it lead to several examples of how to fix this.


petermisik (approved):
Wow.  thank you for the fix.  That is slick.

PR #105, opened 2020-12-16
fix string error on numeric extension.

jbethanco:


jbethanco:
Verified fix for pdf generation crash. Embarrassing.

Pete-Hoch (approved):
Thanks!

PR #106, opened 2020-12-17
Close Bug PR-210 in prep to fix PR-209

jbethanco:
Form data drawing front and back were being called twice because the function that generated the front OR back of form was generating the front AND back of form.
Refactored all the functions in the process of  form drawing and form image generation, to make it easier to write test cases.
The decomposed functions might be useful for other forms later?
Rewrote the print function to use a pdf instead of images and html, so we don't have the extra maintenance but with the same output.
Changed Helper.WIDTH / Helper.HEIGHT to Helper.LETTER_SIZE CGSize to decrease verbosity.

### DIFF
--- a/Logging/Controllers/View Controllers/Form 781/Form781ViewController.swift
+++ b/Logging/Controllers/View Controllers/Form 781/Form781ViewController.swift
@@ -101,7 +101,7 @@ class Form781ViewController: UIViewController {
     }
         
     @IBAction func printButtonTapped(_ sender: UIButton) {
-        Helper.printFormFunc()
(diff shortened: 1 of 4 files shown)


Pete-Hoch:
This looks nice. Both Send & Print dialogs seem about 30% faster to display. Printing is good, but it still has long ends binding. Short edge makes more sense.

Pete-Hoch:
Sounds like you and @petermisik should team up on this.

petermisik (approved):
I'm updating my PR to fix your typo.  Other than that, this is amazing.  Great work.

PR #107, opened 2020-12-17
Put pdf generation on a background thread.  Menu speed improved.

petermisik:


jbethanco:
This is going to conflict with the branch I submitted last night. Could you re-attack after that merges? 
Also, I think it would be a good idea to have a completion call back so that Bethany can manage UI state while processing.

Pete-Hoch (commented):
I used SEND to print and then used PRINT to Print and crashed on the second Print.

Pete-Hoch (approved):
Looks reasonable.

Too bad we can't do this for the Print dialog too.

petermisik:
Do you think we need the print button on the page?

jbethanco (approved):
Looks good. I think print781 will need a completion callback, and the print button should be disabled/reenabled accordingly.

Pete-Hoch:
> Do you think we need the print button on the page?

It depends on how often Print will be used vs anything else. If everyone will use Print, it's a convenience. But if printing is rare, I think we should drop it since it's supported in the share dialog.

PR #108, opened 2020-12-17
Fix/quick fix for flight seq pop up

bmorris-revacomm:
- did a quick fix for the flight seq pop up so that everything is visible in landscape and portrait (will work on fixing UI later)
- connected addDayButton for @Pete-Hoch so he can start working on functionality (tableview outlet should already be there)


petermisik (commented):
Change looks good to me.

jbethanco (approved):
Nice

PR #109, opened 2020-12-17
Change Flying Origin to Flying Organization.

Pete-Hoch:
Note, this change will cause current Forms to not be readable anymore.


jbethanco (approved):
Pulled changes. Tested quick. Looks good. 
Needed to happen. That was confusing.

PR #113, opened 2020-12-18
Fix/tests and alignment

jbethanco:
This fixes the form data alignment for the aircrew section.
Includes two unit tests for form generation validation.
One test includes a boolean, when enabled, saves png images of the form to the local disk for manual inspection. 
I commented the test out for for the second form (1). If someone could test out that test I'd appreciate it. 
It works on my computers here, but doesn't work on the Github server. Curious if it works for everyone else at home.


bmorris-revacomm (approved):
looks good to me!

PR #114, opened 2020-12-18
Fix/bug fixes for flight seq pop up

bmorris-revacomm:
- dim view extended when scrolling
- keyboard bug fixes


Pete-Hoch (approved):
Looks good.

PR #115, opened 2020-12-18
Fix/swipe between flight time and flight conditions

bmorris-revacomm:
(it kept my commits from the last PR again...)

- can now swipe between flight time table view and flight conditions table view (still need to set up cells properly)
- deleted old files and storyboards
- the page view controller code is older code that Pete M used for something else a couple months ago, so I'm not sure if there is anything I'm missing or that I can remove. I haven't done a page view controller before, so I'm still learning!


Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
```diff
@@ -15,9 +15,13 @@ protocol FlightSeqPopUpViewControllerDelegate: class {
 class FlightSeqPopUpViewController: UIViewController {
     
     // MARK: - Outlets
+<<<<<<< HEAD
+        
+=======
     
     @IBOutlet weak var scrollView: UIScrollView!
     
+>>>>>>> daec656a2a85ffebd70df6872d86b1443c312cc7
```
Pete-Hoch, line 24: Merge conflict need to be fixed.

Pete-Hoch (commented):
I cleaned up the merge conflict locally and ran the code.
Flipping between Flight Conditions and Flight Time works nicely, but it's not very discoverable. Should we also have some chevron buttons '<' '>' to do the transition as well as flip?
Running on an Air, some of the text looks really big. Especially the data in Flight Set and the Send icon overlaps the Print text.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
bmorris-revacomm: Oh okay. Do you know why it doesn't tell me that I have merge conflicts sometimes? I'm also getting some weird error when I try to do git pull, so I might need your help with that on Monday

bmorris-revacomm: Looking at the Air right now - apparently there are some weird things going on with it and the 8th gen ipad. I'll look at getting those fixed

bmorris-revacomm: Looks like there is an Xcode bug that screwed up the font in the simulators - it's happened several times before, but it should look okay now. I also added a page control so the user knows they can swipe!

jbethanco (approved):
More red than green... must be good.

PR #116, opened 2020-12-19
Fix/background form generation

jbethanco:
Really want everyones buy in on this one. 
Not as pretty as I would like. 
Puts a popover with a ActivityIndicatorView (spinner) while waiting for form generation. 
Once generations is complete, jumps back to the main thread and puts the UIActivityViewController popover view for sharing. 
There is a slight delay between the popovers changing. 
I'm also not sure if the share popover looks the same as the original, but we can tweak that. 

**Lesson learned:** 
In "DispatchQueue.global(qos: .background).async" 
.background is low priority
Making it "DispatchQueue.global(qos: .userInitiated).async " speeds things up. 

_**QOS Types:**_

- User Interactive: Work that happens on the main thread, such as animations or drawing operations.
- User Initiated: Work that the user kicks off and should yield immediate results. This work must be completed for the user to continue.
- Utility: Work that may take a bit and doesn’t need to finish right away. Analogous to progress bars and importing data.
- Background: This work isn’t visible to the user. Backups, syncs, indexing, etc.


cbrech:
Screen Shot 2020-12-20 at 3 07 14 PM

I like that it has the activity indicator, shows the user that their request has been received and is being processed. My only beaf with it being in the small popover is that you can navigate away while its thinking and then get a share menu attached to nothing.  I think this is really small potatoes not something that needs fixed for 0.1

jbethanco:
> Screen Shot 2020-12-20 at 3 07 14 PM
> 
> I like that it has the activity indicator, shows the user that their request has been received and is being processed. My only beaf with it being in the small popover is that you can navigate away while its thinking and then get a share menu attached to nothing. I think this is really small potatoes not something that needs fixed for 0.1

Yeah... I'm not liking it. Sadly the library provided UIActivityViewController (a.k.a. share stuff view) is expecting that you are ready to share stuff when you decide to share stuff. It provides no way to display one is waiting for the stuff to share to be created after presenting the share stuff view. The share button shouldn't be available if we have nothing ready to share. 

I'm going stop the pull request and trash this branch.

The fundamental problem is form generation taking too long. I've started work on a pdf filler with PDFKit now that we have an unencrypted form. Form generation should take near zero time as it is just filling out text fields. 

Benefits: Faster generation, smaller pdf, editable pdf, signable pdf, more maintainable. 
Drawbacks: My 16 yr old charged me 20 bucks to do the monotonous field naming in the pdf.

cbrech:
@jbethanco I don't think this work needs trashed. I think its a good interim solution!

jbethanco:
> @jbethanco I don't think this work needs trashed. I think its a good interim solution!

LOL. I think it deserves to go in the bin. It's not pretty and it will make it more difficult to implement the proper solution.

PR #118, opened 2020-12-21
Cleaned up and refactored helper functions file

jbethanco:
Refactored and cleaned up items in the HelperFunctions.swift file. 
Made several extension and an extensions folder group

Some of the differences are minor and meant to just help with code reasoning and readability. 

Did unique commits for each change to make it easier to see what was changed and for what reason. 

It's been my experience that  Helper files become a catch all place to put functions when we want to avoid thinking about the overall architecture. I'd recommend in we try to remove it and put the functions in logical name spaces. 

e.g. PDF handling should be in a pdf class, printing should be in a printer class, text field highlighting should be in text field extensions.


petermisik (commented):
Changes look solid to me.  Thanks for the clean up.

petermisik (commented):
I like the changes.  Make sure you add reviews to PRs.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/AircrewPopUpViewController.swift
```diff
@@ -98,19 +98,19 @@ class AircrewPopUpViewController: UIViewController {
...
-        Helper.unhighlight(textField: lastName)
-        Helper.unhighlight(textField: firstName)
-        Helper.unhighlight(textField: ssn)
-        Helper.unhighlight(textField: flightAuthDutyCode)
-        Helper.unhighlight(textField: flyingOranization)
+        lastName.removeHighlight()
+        firstName.removeHighlight()
+        ssn.removeHighlight()
+        flightAuthDutyCode.removeHighlight()
+        flyingOranization.removeHighlight()
```
Pete-Hoch, line 113: Highlighted has been around as a term for a long time in the UI kit. 

Can we name this, removeRedHighlight()?

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
```diff
@@ -134,32 +134,37 @@ class FlightSeqPopUpViewController: UIViewController {
...
+        toICAO.removeHighlight()
+        specialUse.removeHighlight()
+        takeOffTime.removeHighlight()
+        landTime.removeHighlight()
+        totalTime.removeHighlight()
+        touchAndGo.removeHighlight()
+        fullStop.removeHighlight()
+        totalLandings.removeHighlight()
+        sorties.removeHighlight()
+
```
Pete-Hoch, line 167: Ditto.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
```diff
@@ -208,70 +213,70 @@ class FlightSeqPopUpViewController: UIViewController {
...
-            Helper.highlightRed(textField: sender)
+            sender.highlightRed()
         } catch {
             NSLog("checkTakeOffTime function unknown Error")
         }
     }
     
     @IBAction func calculateTotalTime(_ sender: Any) {
-        if Helper().checkInput(time: takeOffTime.text!) {
+        if takeOffTime.text!.isExactlyFourCharacters() {
```
Pete-Hoch, line 233: I think as we tighten up out input validation, we will want to make tests in several places. (Not all data will be coming from the user.) So, we may actually want a validation suite that simply takes in a text string. (Helps with unit tests also.)

For now, this is a simple test for a valid time entry, let's name it so and perhaps add the other checks for valid time all in one place.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
```diff
@@ -208,70 +213,70 @@ class FlightSeqPopUpViewController: UIViewController {
...
     }
         
     @IBAction func calculateTotalLandings(_sender: Any) {
         //Here's where we do the math for filling in the total field
-        if Helper.validateNumericalInput(input: touchAndGo){
-            if Helper.validateNumericalInput(input: fullStop){
-                Helper.unhighlight(textField: touchAndGo)
-                Helper.unhighlight(textField: fullStop)
+        if touchAndGo.text!.isDigits{
+            if fullStop.text!.isDigits{
```
Pete-Hoch, line 270: isPositiveDigits()?

We probably don't validate that, but we should. there are no -1 landings.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/MissionDataPopUpViewController.swift
```diff
@@ -59,7 +59,7 @@ class MissionDataPopUpViewController: UIViewController {
             flightAuthTextField.text = form?.flightAuthNum
             issuingUnitTextField.text = form?.issuingUnit
         } else {
-            dateTextField.text = Helper.getTodaysDate()
+            dateTextField.text = Date().aftoFormattedDate()
```
Pete-Hoch, line 62: getTodaysDate() seems for understandable. What is af? (I thought is stood for as fuck.)

Logging/Extensions/Date+Ext.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by John Bethancourt on 12/20/20.
+//  Copyright © 2020 Christian Brechbuhl. All rights reserved.
+//
+
+import Foundation
+
+extension Date {
+    func aftoFormattedDate() -> String {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = Constants.dateFormat
```
Pete-Hoch, line 14: Please use caps for constants, DATE_FORMAT.

Logging/HelperFiles/Constants.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by John Bethancourt on 12/20/20.
+//  Copyright © 2020 Christian Brechbuhl. All rights reserved.
+//
+
+import UIKit
+
+enum Constants {
+    static let dateFormat = "dd MMM yyyy"
+    static let letterPaperResolution = CGSize(width: 3250, height: 2300)
```
Pete-Hoch, line 13: We should talk about style guidelines, but I think it's fairly common to have constants in all caps.

Logging/HelperFiles/HelperFunctions.swift
```diff
@@ -91,129 +39,62 @@ class Helper {
...
+        
     }
-
+    
     static func vmCalculateTotalTime(takeOffTime: String?, landTime: String?) -> String {
-        // Thought here is to calculate the flying hours.
+        
+        guard let takeOffTime = takeOffTime?.replacingOccurrences(of: ":", with: "") else { return "0" }
+        guard let landTime = landTime?      .replacingOccurrences(of: ":", with: "") else { return "0" }
+        guard takeOffTime.isExactlyFourCharacters(),
```
Pete-Hoch, line 71: I think we need to do a little research here, for example is '3:00' a valid time? Or will we require '03:00'. And will we only be getting this from the UI of the app, or will we get some times in the json download from the server?

LoggingTests/LoggingTests.swift
```diff
@@ -11,45 +11,42 @@ import XCTest
...
+    //This tests a function by duplicating the function it is meant to test
+//    func testGetTodaysDate() {
+//        let date = Helper.getTodaysDate()
+//
+//        let dateFormatter = DateFormatter()
+//        dateFormatter.dateFormat = Constants.dateFormat
+//        let testDate = dateFormatter.string(from: Date())
+//
+//        XCTAssertEqual(date, testDate)
+//    }
```
Pete-Hoch, line 24: Why is this not a valid test? And if it is, how do you suggest we test that getTodaysDate() actually returns today's date? Or should we not test for that?

LoggingTests/LoggingTests.swift
```diff
@@ -11,45 +11,42 @@ import XCTest
...
+//
+//        let path = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).last
+//        let url = path?.appendingPathComponent("Logging.json", isDirectory: false)
+//        var result = Helper.doesFileExist(atURL: url!)
+//        if !result {
+//            Form781Controller.shared.create(date: Helper.getTodaysDate(), mds: "C017A", serialNumber: "90-0534", unitCharged: "437 AW (AMC) /DKFX", harmLocation: "JB CHARLESTON SC 29404", flightAuthNum: "20-0772", issuingUnit: "0016AS")
+//            result = Helper.doesFileExist(atURL: url!)
+//        }
+//        XCTAssertTrue(result)
+//    }
```
Pete-Hoch, line 37: Then lets rename the test from testFileFound() (Not testFileNOTFound()) to testFileCreation() and leave the test in. Please to not remove unit tests.

LoggingTests/LoggingTests.swift
```diff
@@ -11,45 +11,42 @@ import XCTest
...
-    func testSeperateHours() {
-        // Test requires input to have a colon
-        let hrs = Helper.separateHoursAndMins(strInput: "10:00", pointer: "hour")
-        XCTAssertEqual(hrs, "10")
+    func testStringExtensions() {
+        XCTAssertTrue("0900".isExactlyFourCharacters())
+        XCTAssertFalse("090".isExactlyFourCharacters())
+        
+        XCTAssertTrue("0123".isDigits)
+        XCTAssertFalse("123A".isDigits)
```
Pete-Hoch, line 45: So we are not testing the "00:00" format anymore? This test was supposedly written for a reason. Please, please, please, DO NOT remove unit tests unless the test is agreed to be invalid.

LoggingTests/LoggingTests.swift
```diff
@@ -11,45 +11,42 @@ import XCTest
...
+        
     }
     
-    func testSeperateMins() {
-        // Test requires input to have a colon
-        let min = Helper.separateHoursAndMins(strInput: "10:30", pointer: "min")
-        XCTAssertEqual(min, "30")
-    }
```
Pete-Hoch, line 52: Ditto. Removing tests is bad unless there is an obvious reason. This test was written for a reason. What is the reason for deleting it?

Pete-Hoch (changes requested):

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
jbethanco: Right on. This was just refactoring what existed that checked for exactly four characters. I thought a more generic name that said exactly what it does was appropriate as an extension name as it can be used on other strings for other cases. Thoughts?

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
jbethanco: This was replacement extension for a function. Trying to keep it focused on the task at hand. Checking for negative would be another test after converted to Int if needed.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/MissionDataPopUpViewController.swift
jbethanco: AFTO Formatted Date 
Date() gives us todays date. 
getTodaysDate gave us a AFTO Formatted string representation of todays date.
Date().aftoFormattedDate() does the same
Should It be AFTOFormattedString()

Logging/Extensions/Date+Ext.swift
jbethanco: I think this is a style guide issue. 
All caps snake case is an objective c / old school remnant 
Modern swift style guides use lower case with enum namespacing
...
https://google.github.io/swift/#global-constants
https://github.com/raywenderlich/swift-style-guide#constants
https://github.com/airbnb/swift#namespace-using-enums

Logging/HelperFiles/Constants.swift
jbethanco: ""

Logging/HelperFiles/HelperFunctions.swift
jbethanco: Agreed. Hopefully the server will use standard ISO 8601 date formats, as should we for sending. 
Our 781 object is stringified and needs to use appropriate data types. We are reinventing the wheel for validations and math in a lot of functions.

LoggingTests/LoggingTests.swift
jbethanco: I'm not certain which test this was, but I think it It was because the code used to test the code is the code being tested. 

e.g. 
doTheThing() -> String{
   return generateTheString()
}

testDoTheThing() {
   XCAssertEquals(doTheThing(), generateTheString())
}

LoggingTests/LoggingTests.swift
jbethanco: It is testing functions of the STL and nothing specific to our app. 
Or am I missing something?

LoggingTests/LoggingTests.swift
jbethanco: The seperateHoursAndMins function was removed, so no longer needs a test. 
The function using that now just removes the colon and continues with a 4 digit string. 

I'm not removing anything until is agreed. This is a reviewable pull request for this purpose.

LoggingTests/LoggingTests.swift
jbethanco: The function no longer exists. It can't be tested.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/AircrewPopUpViewController.swift
jbethanco: It will remove any color highlight and return the textField to our apps normal design state.

Pete-Hoch: OK, well this is used to undo the addRedHightlightIfBlank() functions. The calls seem a little asymmetrical. How about?

addRedHightlightIfBlank() & removeAnyColorHighlithg() or
addColorHighlightIfBlank(color:) & removeColorHilight()?

jbethanco: Sounds good.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
Pete-Hoch: Honestly, when we get to validation (which we hardly do) I think that we will end up with very data specific code. So a Time validation seems more reasonable. Names are unlikely to have any validation, but Flight Auth # might. Issuing Unit might also, etc. 

Using 4 characters validates what, using time validates what and why.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/MissionDataPopUpViewController.swift
Pete-Hoch: So, maybe a comment or something that explains what AFTO stands for would be helpful here.

Logging/HelperFiles/HelperFunctions.swift
Pete-Hoch: Mu concern is that (I believe) 03:34 would have been understood at a time before this change and after the change it will not.

LoggingTests/LoggingTests.swift
Pete-Hoch: Yes, so if someone changed the code in the function this test would break. Now if someone makes a change and getTodaysDate() no longer returns todays date, we have a bug that is not caught by the unit tests.

So I ask again, how do you suggest we test to make sure that getTodaysDate() works correctly?

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/MissionDataPopUpViewController.swift
jbethanco: yeah. or perhaps I could name it more clearly...
Date().AFTOFormFormatString()
with similar casing "to testHTTPConnection"

LoggingTests/LoggingTests.swift
Pete-Hoch: it's testing that we are using the STL correctly. And it's also testing for when something changes in the future.

LoggingTests/LoggingTests.swift
Pete-Hoch: Yes, I'm a little concerned that we are not as thorough handling of ##:## anymore, but I guess if things start to silently fail, we will figure it out.

LoggingTests/LoggingTests.swift
jbethanco: getTodaysDate has been removed so we don't need a test there. 
We get the date with STL's Date()
We then format it with Date.AFTOFormatedDate()
We will then need to test that the AFTO Formatted date is coming out correctly. 
I'll add a test for that.

LoggingTests/LoggingTests.swift
jbethanco: I'll write a test for the func doesFileExist(atURL url: URL) -> Bool 
to replace this test

Pete-Hoch: Actually, this is testing to see if Form781Controller.shared.create() actually creates the storage file.

I think it's kind of important.

LoggingTests/LoggingTests.swift
jbethanco: It's being handled by removing the ":" from the string. The UI is also handling this as it doens't allow it.

PR #119, opened 2020-12-21
Fix/finished flight time UI

bmorris-revacomm:
- finished basic UI (some details still need to be added)
- connected all outlets
- crew members now load on both table views


Pete-Hoch (approved):
Looks nice, let's get it in.

PR #122, opened 2020-12-22
Feature/pdf filler

jbethanco:
Uses PDFKit to fill a fillable version of the AFTO Form 781.
And... still not fast. 
The bottleneck appears to be when PDFKit converts the PDFDocument to a data representation. 
Shaved a second or two off by not messing with blank fields. 
Want to talk to Christian and see if having the form fillable upon receipt would be a nicety. If not, we could remove all the unused fields which I think would speed up conversion time. Need to test that hypothesis. 
Either way, this is still faster than before. The delay on share or print is noticeable but not as pronounced. 
-Share to Mattermost works ( I believe it has to do with what you share in the activity popover. The URL to the pdf [works] verses the data representation of the pdf [doesn't work])


Pete-Hoch:
I do not have a printer selected and in the Print dialog, it dismisses when tapping on anything. So I am not able to select the printer, or change the page count, etc.

Using the shared menu crashed once, but I have not been able to reproduce it.

Pete-Hoch (commented):
I like the changes, and even though it's not really faster, it is a whole lot less fragile. But we need to figure out what is going on with the print dialog.

Pete-Hoch:
I used the Share dialog to save to files, but there is nothing in my local files on the iPad. 

(This has all been in the simulator.)

jbethanco:
Try doing a clean build. It's working here. Either way, think we should merge as this is the way going forward.

jbethanco:
> I like the changes, and even though it's not really faster, it is a whole lot less fragile. But we need to figure out what is going on with the print dialog.

Just tested on 5 different iPad simulators and a physical first gen iPad Pro. All work. 

For saving... make sure you are going to "Files-> On my iPad"

Screen Shot 2020-12-22 at 8 36 27 AM
Screen Shot 2020-12-22 at 8 35 34 AM

PR #123, opened 2020-12-22
Fix/updated NSLogs in Form781VC

bmorris-revacomm:
- updated NSLogs because the file name was changed, but NSLog file names were not updated


Pete-Hoch (changes requested):
Please remove the xcuserdatad file from the PR.

The swift changes are good.

PR #124, opened 2020-12-22
Support adding a new form to the app.

Pete-Hoch:
Several changes here.

1. Current form is now an index. We always set it to the last form in the list (be default) but we probably want to persist it.
2. Can now get the form date string by index to support the Days array in the sidebar.
3. Since the + button makes a new (mostly) blank form with todays date, we are not adding a blank form (if no exist) so we do not need to worry about the 'no form' situation.
4. The '+' button on the sidebar adds a new form and copies forward the expected data from the previously newest form.
5. The sidebar shows the date of all forms in the list and allows the user to select any form to work on.


bmorris-revacomm (commented):
something minor, but you put "flightDataChange" instead of "flightDataChanged"

Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
```diff
@@ -27,6 +27,13 @@ class FlightTimeTableViewCell: UITableViewCell {
     
     func setUpViews(crewMember: CrewMember) {
         name.text = crewMember.lastName
+        primary.text = crewMember.primary
+        secondary.text = crewMember.secondary
+        instructor.text = crewMember.instructor
+        evaluator.text = crewMember.evaluator
+        other.text = crewMember.other
+        time.text = crewMember.time
```
petermisik, line 35: I have a change ready for this.  This needs to be calculated based on the other cells.

petermisik (approved):
Looks good.

jbethanco:
When I click + under the date nothing visually happens, but it says:

2020-12-22 11:36:12.677075-1000 Logging[6265:471871] Populated flights
2020-12-22 11:36:12.677275-1000 Logging[6265:471871] Populated crew members
2020-12-22 11:36:12.678593-1000 Logging[6265:471871] Created form

jbethanco:
When I restart the app, it updates the sidebar with the additional forms... for the same day. 
I think when the button is pushed, the list should update with the new forms. 
Sounds like a job for diffableDataSource

Pete-Hoch:
> When I restart the app, it updates the sidebar with the additional forms... for the same day.
> I think when the button is pushed, the list should update with the new forms.
> Sounds like a job for diffableDataSource

Did you try deleting DerivedData?

jbethanco:
> > When I restart the app, it updates the sidebar with the additional forms... for the same day.
> > I think when the button is pushed, the list should update with the new forms.
> > Sounds like a job for diffableDataSource
> 
> Did you try deleting DerivedData?

Yes. I don't see the code for updating the list. 

When you tap plus it adds forms (unbeknownst to the user) 
When you close and reopen the app... it updates with all the added forms. 
When you select different forms on the sidebar it doesn't update the displayed form.

Pete-Hoch:
> > > When I restart the app, it updates the sidebar with the additional forms... for the same day.
> > > I think when the button is pushed, the list should update with the new forms.
> > > Sounds like a job for diffableDataSource
> > 
> > 
> > Did you try deleting DerivedData?
> 
> Yes. I don't see the code for updating the list.
> 
> When you tap plus it adds forms (unbeknownst to the user)
> When you close and reopen the app... it updates with all the added forms.
> When you select different forms on the sidebar it doesn't update the displayed form.

Double check that you have the current code. That is all working here.

Pete-Hoch:
@jbethanco 

If you do not have this bit of code in SideMenuViewController then you need to pull to update your source.

        NotificationCenter.default.addObserver(self,
                                               selector: #selector(currentFormChanged),
                                               name: Form781Controller.flightDataChanged,
                                               object: nil)

Pete-Hoch:
Screen Shot 2020-12-22 at 12 17 36 PM

Screen Shot 2020-12-22 at 12 18 01 PM

bmorris-revacomm:
"in the butt bob" hahaha

jbethanco:
> > > > When I restart the app, it updates the sidebar with the additional forms... for the same day.
> > > > I think when the button is pushed, the list should update with the new forms.
> > > > Sounds like a job for diffableDataSource
> > > 
> > > 
> > > Did you try deleting DerivedData?
> > 
> > 
> > Yes. I don't see the code for updating the list.
> > When you tap plus it adds forms (unbeknownst to the user)
> > When you close and reopen the app... it updates with all the added forms.
> > When you select different forms on the sidebar it doesn't update the displayed form.
> 
> Double check that you have the current code. That is all working here.

Was a commit or two behind... all works now.

jbethanco (approved):
It works. What happens if they have 300 forms?

PR #127, opened 2020-12-22
Fix/calculate aircrew time

petermisik:
Updated HelperFunctions with the math to sum the row.


Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
```diff
@@ -32,8 +32,9 @@ class FlightTimeTableViewCell: UITableViewCell {
         instructor.text = crewMember.instructor
         evaluator.text = crewMember.evaluator
         other.text = crewMember.other
-        time.text = crewMember.time
```
Pete-Hoch, line 35: It looks like we never use this anymore. (But we do set it in the constructor.)

Should we remove the field from CrewMember?

Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
```diff
@@ -32,8 +32,9 @@ class FlightTimeTableViewCell: UITableViewCell {
...
+        time.text = Helper.airCrewTotalTimeCalculation(crewMember: crewMember)
```
Pete-Hoch, line 35: How about making this a function in CrewMember? Something like:

crewMember.totalAirTime().

Pete-Hoch (commented):
Just a couple of questions. (And suggestions.)

Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
jbethanco: I'm having a hard time getting past our data model using strings for what should be doubles, ints, or dates. Have we thought about refactoring this? Converting from and to strings to do math isn't ideal.

Pete-Hoch: > I'm having a hard time getting past our data model using strings for what should be doubles, ints, or dates. Have we thought about refactoring this? Converting from and to strings to do math isn't ideal.

Understood, but we are likely to change persistence over to Core Data (or something) after 0.1.0 and we will be able to do that all at once. We can even add formatters to the text fields so the user can't even type a character in a numeric field. But again, this has all been very quickly put together and there is little reason to put a lot of time into the current UI when we plan to totally replace it by mid January.

At that point, we can address the data model and what makes the most sense.

jbethanco: Sounds good!

Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
petermisik: I don't think we need to remove it from CrewMember.  This data needs to be attached to the model and part of the storage.

Logging/Controllers/View Controllers/Split Time/FlightTimeTableViewCell.swift
petermisik: Changes are in the new PR.

jbethanco (changes requested):
Looks good other than the tests needing an update. 
And just FYI: I believe Pete H is wanting two reviews prior to merges till 0.1.0

Pete-Hoch (changes requested):
Need to fix up unit tests.

jbethanco (approved):

Pete-Hoch (approved):
Looks good!

PR #128, opened 2020-12-22
Fix/enabled flight conditions functionality

bmorris-revacomm:
- enabled editing/saving for flight conditions
- updated some UI
- removed editing for time in flight time


Pete-Hoch (approved):
Looks good.

PR #129, opened 2020-12-23
Fix/UI updates

bmorris-revacomm:
- set delegates for textfields on flight time and flight condition
- updated textfield stackviews so numbers are not getting cut off
- removed unnecessary UI elements/keyboard notification


Pete-Hoch (approved):
This is good.

PR #132, opened 2020-12-23
Update Form781+PDF.swift

jbethanco:
Add print to print info:
printInfo.duplex = .shortEdge


Pete-Hoch (commented):
Unfortunately this makes no difference for my printer.

HP Laser.

jbethanco:
> Unfortunately this makes no difference for my printer.
> 
> HP Laser.

Does it need to be long edge? I'm not sure what output you are expecting.

Pete-Hoch:
> > Unfortunately this makes no difference for my printer.
> > HP Laser.
> 
> Does it need to be long edge? I'm not sure what output you are expecting.

I get long edge binding with or without this change.
I also get long edge binding if I set my orientation to landscape.
Everything is long edge binding.

petermisik (commented):
I get the same.  We're looking for short edge because you would naturally turn the form (left) on the short edge and have all of the images facing the same way.

jbethanco:
I'm thinking this may be pdf issue. 
However, our pdf prints the same duplex way as the official version. (if you flip on the short edge, then form upside down)

PR #133, opened 2020-12-23
Change icons for print and shared.

jbethanco:
Looks like: 
Screen Shot 2020-12-23 at 12 58 46 PM


Pete-Hoch (approved):
Looks good.

Pete-Hoch:
I've reviewed this twice, ship it!

PR #134, opened 2020-12-24
Fix/refactor rename comment clean

jbethanco:
Made some name changes.
Cleaned up some functions.
Reduced some lines.
Added clarity. 
Made some comments.
Updated call sites and tests to match.


Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
```diff
@@ -212,7 +212,7 @@ class FlightSeqPopUpViewController: UIViewController {
     
     @IBAction func checkTime(_ sender: UITextField) {
         do {
-            let _ = try Helper.validateTime(timeString: sender.text!)
+            let _ = try Utilities.validateTime(sender.text!)
```
Pete-Hoch, line 215: This may seem like a more logical name to you, but this is what everyone else has been using for months. Why is Utilities better than Helper? It's more characters to type. This seems like a change for change sake. Perhaps we should discuss this as a team?

Logging/HelperFiles/Utilities.swift
```diff
@@ -0,0 +1,137 @@
+//
+//  FirstVCViewModel.swift
+//  Logging
+//
```
Pete-Hoch, line 4: I'm not sure what was changed in this file because there is also a name change of the file. But the header should at least match the name of the file now.

Logging/Extensions/String+Ext.swift
```diff
@@ -8,8 +8,21 @@
 
 import Foundation
 
+infix operator +^+
+
```
Pete-Hoch, line 12: Why are we spending time to support Strings that are really Ints when we want to change the data model to eliminate this?

LoggingTests/LoggingTests.swift
```diff
@@ -69,50 +72,50 @@ class LoggingTests: XCTestCase {
...
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: nil, landTime: "2300"), "0")
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: "0000", landTime: "0005"), "0.1")
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: "0005", landTime: "2300"), "22.9")
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: "2301", landTime: "00:01"), "1.0")
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: "07:30", landTime: "14:35"), "7.1")
-        XCTAssertEqual(Helper.vmCalculateTotalTime(takeOffTime: "20:12", landTime: "07:20"), "11.1")
+    func testVmCalculateTotalTime(){
+        XCTAssertEqual(Utilities.timeBetween(nil, "2300"), "0")
+        XCTAssertEqual(Utilities.timeBetween("0000", "0005"), "0.1")
+        XCTAssertEqual(Utilities.timeBetween("0005", "2300"), "22.9")
```
Pete-Hoch, line 78: I personally do not find this clearer at all.

Pete-Hoch (commented):
First I don't see anything we need for 0.1.0 here.

Also, I do not see the bug in printing you talk about. Printing should be using the currentForm() function.

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
petermisik: I'm not seeing the need to change the file.

LoggingTests/LoggingTests.swift
jbethanco: I know we aren't merging due to our offline talk. But curious... what would you suggest as a clear function name for this?

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
jbethanco: - My thinking was every function in an app "helps" an app, so Helpers tends to be a catch all. 
- This is previously where we had pdf generation, pdf printing, pdf sharing, math, and field checking. 
- Utilities has a slightly more focused meaning and from my experience, the typical name for this type of class/struct. 
- My personal preference would be to eliminate this type of file and put functions in appropriate namespaces or as extensions on the effected models.

Logging/Extensions/String+Ext.swift
jbethanco: Because it is still code that is visible to others. :)

LoggingTests/LoggingTests.swift
Pete-Hoch: I like calculate when something is being calculated, and the result is a duration (of time) so I would probably start with calculateDuration(), I would also put the labels in. Swift has labels for (what I think) are good reasons. So I wold propose this when starting over:

func calculateDuration(startTime: Date, endTime: Date) -> Double

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/FlightSeqPopUpViewController.swift
Pete-Hoch: Yea, let's see if we can pt functions in associated classes and do our best to not have helper or utility buckets.

(Maybe just a Utility group (folder) in the project to put swift extensions, etc.)

PR #136, opened 2020-12-28
Fix/keyboard bug

bmorris-revacomm:
- if the keyboard is up when a pop up is closed, it will now be dismissed


Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/AircrewPopUpViewController.swift
```diff
@@ -128,12 +128,21 @@ class AircrewPopUpViewController: UIViewController {
         flightAuthDutyCode.text = crewMember.flightAuthDutyCode
         flyingOranization.text = crewMember.flyingOranization
     }
+    
+    func dismissKeyboard() {
+        firstName.resignFirstResponder()
+        lastName.resignFirstResponder()
+        ssn.resignFirstResponder()
+        flyingOranization.resignFirstResponder()
+        flightAuthDutyCode.resignFirstResponder()
```
Pete-Hoch, line 137: This is very brute force. There is only one first responder at any time. So you only have to resign first responder on the one that is currently editing.

There are several ways to do this, but there is a nice extension to IUView that helps:

self.view.endEditing(true)

endEditing(_ force: Bool) will end editing on and subview of the Vire and ending editing will resign the responder (as well as other things).

This also means that if you have a common class for the editing view controllers, you only have to write this method once. (Ore replace dismissKeyboard() with the single line of code.)

Pete-Hoch (approved):
Try out my suggestion and then either change it or not before merging into release/0.1.0

Logging/Controllers/View Controllers/Form 781/Form781 Pop Ups/AircrewPopUpViewController.swift
bmorris-revacomm: Yes, it worked like a charm! Thank you!! I am going to close this PR because I sent out another one with the changes

PR #138, opened 2021-01-01
Just the basic app project and the Header view.

Pete-Hoch:
Nothing is hooked up to do anything.


jbethanco (commented):
Thought we were going with the "PBL" name vs "Logging" to avoid namespace confusion with logging frameworks. Time for fireworks.

Pete-Hoch:
> Thought we were going with the "PBL" name vs "Logging" to avoid namespace confusion with logging frameworks. Time for fireworks.

When I talked about using PBL, it was in the context of having both the 0.1.0 code and the 0.2.0 code in the same branch of the repository. But the 0.1.0 branch was archived and pushed to 0.1.1 and the 0.2.0 branch was started clean. So, I'm not sure if there is a conflict risk. But I'm happy to use the PBL prefix. I think we need to keep the application name as "Logging" to make life easier for everyone consuming it. But it's probably a good idea to have PBLContentView, etc. (Is that what you mean?)

Pete-Hoch:
We are failing to build due to a lack of signing. The team in 0.1.0 is set to "RQ9T47YT3S", but there is also an error message there that this is an unknown team. And I do not seem to be able to add the signing to the project.

@cbrech, is this something you need to do?

jbethanco:
> > Thought we were going with the "PBL" name vs "Logging" to avoid namespace confusion with logging frameworks. Time for fireworks.
> 
> When I talked about using PBL, it was in the context of having both the 0.1.0 code and the 0.2.0 code in the same branch of the repository. But the 0.1.0 branch was archived and pushed to 0.1.1 and the 0.2.0 branch was started clean. So, I'm not sure if there is a conflict risk. But I'm happy to use the PBL prefix. I think we need to keep the application name as "Logging" to make life easier for everyone consuming it. But it's probably a good idea to have PBLContentView, etc. (Is that what you mean?)

When I was discussing it, I was concerned with conflicts with Apple's logging framework or even OS Log if we ever go that route. e.g.
import Logging 
https://github.com/apple/swift-log

Might cause confusion and conflicts. I was initially confused with the current naming and was trying to find the logging framework and how it was being used. 

I think PBLogging or PBL as a project name is safer, and also PBL as a prefix for our custom subclasses of Apples classes. 

But if we are going to live dangerously... let's go all the way and call our project "UIKit."  :) 

Hope you had a good new years eve.

cbrech:
@Pete-Hoch I fixed the code signing, have to take it off automatic for release so the pipeline can use its provisioning profile.

cbrech (changes requested):
suggest adding the build script that generates the build ID from the git commit before merging into release/2.0. Otherwise that pipeline will fail trying to upload to Apple.

Pete-Hoch:
> suggest adding the build script that generates the build ID from the git commit before merging into release/2.0. Otherwise that pipeline will fail trying to upload to Apple.

Good catch! Added script.

cbrech (approved):

Pete-Hoch:
I'm a little hesitant to change the project name because of all the scripts we have. Also I am not sure how it ties into the rest of the system.

I believe we need to leave the app as "Logger" but maybe now.

What does everyone else think?

cbrech:
> I'm a little hesitant to change the project name because of all the scripts we have. Also I am not sure how it ties into the rest of the system.
> 
> I believe we need to leave the app as "Logger" but maybe now.
> 
> What does everyone else think?

I think the project name can change, the bundle ID needs to stay though

Pete-Hoch:
I think this is a reasonable starting place. Let me know if you want any more changes before this lands on release/0.2.0.

jbethanco:
> I'm a little hesitant to change the project name because of all the scripts we have. Also I am not sure how it ties into the rest of the system.
> 
> I believe we need to leave the app as "Logger" but maybe now.
> 
> What does everyone else think?

Logger? I think you meant Logging, but Logger would be different enough from the Apple stuff if that is what you meant. I think as long as the bundle ID is the only thing that matters for the scripts we should go with PBLogging or PBL. We'd probably be fine with Logging, though I'd rather not risk it since the cost to change now is near-free and the potential cost later would be greater.

Pete-Hoch:
> Logger? I think you meant Logging, but Logger would be different enough from the Apple stuff if that is what you meant. I think as long as the bundle ID is the only thing that matters for the scripts we should go with PBLogging or PBL. We'd probably be fine with Logging, though I'd rather not risk it since the cost to change now is near-free and the potential cost later would be greater.

Ha! Yes, typo. I think of it as Logger some times, but the app name (so far) is Logging. So I have change project names with Logging to start with PBLogging and added PBL in front of our source files. (But I did not mess with things like AppDelegate. (Although I have no problem changing the name to PBLAppDelegate. It's just a bit of a PITA.)

jbethanco:
> > Logger? I think you meant Logging, but Logger would be different enough from the Apple stuff if that is what you meant. I think as long as the bundle ID is the only thing that matters for the scripts we should go with PBLogging or PBL. We'd probably be fine with Logging, though I'd rather not risk it since the cost to change now is near-free and the potential cost later would be greater.
> 
> Ha! Yes, typo. I think of it as Logger some times, but the app name (so far) is Logging. So I have change project names with Logging to start with PBLogging and added PBL in front of our source files. (But I did not mess with things like AppDelegate. (Although I have no problem changing the name to PBLAppDelegate. It's just a bit of a PITA.)

Right on. I'm not a fan of overusing the prefix either. Just in situations where we need to deconflict with Apple or a framework. Normally when subclassing or mirroring something else. e.g. PBLAlertView, PBLResults, or PBLColor.

Pete-Hoch:
> > > Logger? I think you meant Logging, but Logger would be different enough from the Apple stuff if that is what you meant. I think as long as the bundle ID is the only thing that matters for the scripts we should go with PBLogging or PBL. We'd probably be fine with Logging, though I'd rather not risk it since the cost to change now is near-free and the potential cost later would be greater.
> > 
> > 
> > Ha! Yes, typo. I think of it as Logger some times, but the app name (so far) is Logging. So I have change project names with Logging to start with PBLogging and added PBL in front of our source files. (But I did not mess with things like AppDelegate. (Although I have no problem changing the name to PBLAppDelegate. It's just a bit of a PITA.)
> 
> Right on. I'm not a fan of overusing the prefix either. Just in situations where we need to deconflict with Apple or a framework. Normally when subclassing or mirroring something else. e.g. PBLAlertView, PBLResults, or PBLColor.

Well, we have PBLogging now because Logging confided you, but no-one else (at least no-one else has said so) and in my understanding prefixing is used to avoid a naming conflict both currently and in the future. So I think we should use it on all of our files just to make the distinction clear. And then if Apple adds a MainView (as a random example) we will already have a PBLMainView and so there will not be a conflict.

This is also a good habit to get into getting ready for when we have our own library that are used in different applications.

jbethanco:
> > > > Logger? I think you meant Logging, but Logger would be different enough from the Apple stuff if that is what you meant. I think as long as the bundle ID is the only thing that matters for the scripts we should go with PBLogging or PBL. We'd probably be fine with Logging, though I'd rather not risk it since the cost to change now is near-free and the potential cost later would be greater.
> > > 
> > > 
> > > Ha! Yes, typo. I think of it as Logger some times, but the app name (so far) is Logging. So I have change project names with Logging to start with PBLogging and added PBL in front of our source files. (But I did not mess with things like AppDelegate. (Although I have no problem changing the name to PBLAppDelegate. It's just a bit of a PITA.)
> > 
> > 
> > Right on. I'm not a fan of overusing the prefix either. Just in situations where we need to deconflict with Apple or a framework. Normally when subclassing or mirroring something else. e.g. PBLAlertView, PBLResults, or PBLColor.
> 
> Well, we have PBLogging now because Logging confided you, but no-one else (at least no-one else has said so) and in my understanding prefixing is used to avoid a naming conflict both currently and in the future. So I think we should use it on all of our files just to make the distinction clear. And then if Apple adds a MainView (as a random example) we will already have a PBLMainView and so there will not be a conflict.
> 
> This is also a good habit to get into getting ready for when we have our own library that are used in different applications.

I think our Objective C roots are showing and are hard to shake. We may be overthinking this. Swift has namespaces that avoid collisions making these type of prefixes rarely necessary. The Apple recommended 3 letter class prefixes were for Objective C. If we make our own package it will have its own namespace. Module names are a different story. If we take a look at a few of the apple provided swift projects like Fruta or newer framework like SwiftUI... it seems they've eliminated the prefix practice all together. I just took a look at Ray Wenderlichs style guide... He may have summed up the issue nicely:

"Swift types are automatically namespaced by the module that contains them and you should not add a class prefix such as RW. If two names from different modules collide you can disambiguate by prefixing the type name with the module name. However, only specify the module name when there is possibility for confusion, which should be rare."

I'm finding recently that a lot of my Objective C habits and best practices don't make sense. Really have to look back at the reasoning behind them and see if they are necessary anymore.

Pete-Hoch:
So get rid of all of the prefixing?

jbethanco:
> So get rid of all of the prefixing?

I can't think of a reason to keep it on the classes.

Pete-Hoch:
> > So get rid of all of the prefixing?
> 
> I can't think of a reason to keep it on the classes.

Well if the reason to have a prefix (or not) is unavoidable name collisions, then I can't see a reason for having them anywhere. 

Saying, we only use the prefix for a name collision is very straight forward. Anything else seems like personal preference.

jbethanco:
> > > So get rid of all of the prefixing?
> > 
> > 
> > I can't think of a reason to keep it on the classes.
> 
> Well if the reason to have a prefix (or not) is unavoidable name collisions, then I can't see a reason for having them anywhere.
> 
> Saying, we only use the prefix for a name collision is very straight forward. Anything else seems like personal preference.

It seems collisions on classes can be avoided with namespaces. Collisions on module names can't as they are the namespace.  
In a lot of projects I've seen recently they use prefixes for app specific styling  (PBLButton) or highlighting that a namespace is app specific (PBLResult.fileNotFound).

Pete-Hoch:
> > > > So get rid of all of the prefixing?
> > > 
> > > 
> > > I can't think of a reason to keep it on the classes.
> > 
> > 
> > Well if the reason to have a prefix (or not) is unavoidable name collisions, then I can't see a reason for having them anywhere.
> > Saying, we only use the prefix for a name collision is very straight forward. Anything else seems like personal preference.
> 
> It seems collisions on classes can be avoided with namespaces. Collisions on module names can't as they are the namespace.
> In a lot of projects I've seen recently they use prefixes for app specific styling (PBLButton) or highlighting that a namespace is app specific (PBLResult.fileNotFound).

Reading this, it sounds like you are saying to remove all prefixing (until we need it). But that can't be what you mean, because that is where we started. For example, you say we can't avoid a collision on module names using namespaces, but that is exactly what you said above that we should do, IF THERE IS A COLLISION.

So please, instead of me trying to read your mind and changing this project back and forth to still have you unsatisfied, can you just change the names to whatever it is you want and the rest of us will try and follow your lead.

jbethanco:
> > > > > So get rid of all of the prefixing?
> > > > 
> > > > 
> > > > I can't think of a reason to keep it on the classes.
> > > 
> > > 
> > > Well if the reason to have a prefix (or not) is unavoidable name collisions, then I can't see a reason for having them anywhere.
> > > Saying, we only use the prefix for a name collision is very straight forward. Anything else seems like personal preference.
> > 
> > 
> > It seems collisions on classes can be avoided with namespaces. Collisions on module names can't as they are the namespace.
> > In a lot of projects I've seen recently they use prefixes for app specific styling (PBLButton) or highlighting that a namespace is app specific (PBLResult.fileNotFound).
> 
> Reading this, it sounds like you are saying to remove all prefixing (until we need it). But that can't be what you mean, because that is where we started. For example, you say we can't avoid a collision on module names using namespaces, but that is exactly what you said above that we should do, IF THERE IS A COLLISION.
> 
> So please, instead of me trying to read your mind and changing this project back and forth to still have you unsatisfied, can you just change the names to whatever it is you want and the rest of us will try and follow your lead. 



> > > > > So get rid of all of the prefixing?
> > > > 
> > > > 
> > > > I can't think of a reason to keep it on the classes.
> > > 
> > > 
> > > Well if the reason to have a prefix (or not) is unavoidable name collisions, then I can't see a reason for having them anywhere.
> > > Saying, we only use the prefix for a name collision is very straight forward. Anything else seems like personal preference.
> > 
> > 
> > It seems collisions on classes can be avoided with namespaces. Collisions on module names can't as they are the namespace.
> > In a lot of projects I've seen recently they use prefixes for app specific styling (PBLButton) or highlighting that a namespace is app specific (PBLResult.fileNotFound).
> 
> Reading this, it sounds like you are saying to remove all prefixing (until we need it). But that can't be what you mean, because that is where we started. For example, you say we can't avoid a collision on module names using namespaces, but that is exactly what you said above that we should do, IF THERE IS A COLLISION.
> 
> So please, instead of me trying to read your mind and changing this project back and forth to still have you unsatisfied, can you just change the names to whatever it is you want and the rest of us will try and follow your lead.

Sorry I think we are missing wavelengths. I tried to contact you on mattermost a few hours ago so we could voice chat. I'm saying they should be removed on all Classes as they are unnecessary because module namespaces avoid collisions for us. So they are just extra typing. I'm not near my dev computer, so can't make the changes. I hope this isn't coming off as a personal preference, but rather keeping the project normalized with standard practices / style guides. If we aren't on the same page with a change, let's not make it till we are. I'll be back home in an hour. we can talk.

PR #140, opened 2021-01-04
Looks like the launch screen has to be at the top level.

Pete-Hoch:
OK, I think I have it this time.

(Can't drop iOS 13 support soon enough.)


cbrech (approved):

petermisik:
Why are we using a storyboard in swiftUI?  I thought one of the main reasons was to get away from using them.

jbethanco:
Just for launch on iOS 13.

cbrech:
@petermisik the launch/splash screen on iOS 13. Not any of the actual app.

petermisik:
Thank you for the clarification

petermisik (commented):
Nice and clean.

PR #143, opened 2021-01-05
make side view collapse

Pete-Hoch:
Add our chevron and make side view have collapsed and non-collapsed drawing states.


jbethanco (changes requested):
Animation looks strange in simulator and on device. 
I think we should discuss not using NavigationView and trying recreate the wheel.

jbethanco:
> Animation looks strange in simulator and on device.
> I think we should discuss not using NavigationView and trying recreate the wheel.


https://user-images.githubusercontent.com/76016626/103703877-6f660d80-4f4c-11eb-8fdd-e7e7d1c9f7b6.mov

bmorris-revacomm:
Yeah, we'll have to take a look at the animation, but other than that, I think it looks good! Worst-case scenario, we just remove the animation?

jbethanco:
> Yeah, we'll have to take a look at the animation, but other than that, I think it looks good! Worst-case scenario, we just remove the animation?

We need to make it look smooth. We can fix the animation by making the buttons look like this: 
Button(action:{
                        withAnimation{
                            self.isCollapsed.toggle()
                        }
                    }){
                        Image(systemName:"chevron.right")
                    } .padding()
                    
We wrap the boolean toggle with "withAnimation" SwiftUI takes care of animating anything associated with it. Need to do this in both places.  
I think it would be better to have a single button. It can animate the chevron rotation as well. 
Then getting rid of the animation call at the end of the view.

Bigger picture though, we are giving up a ton of functionality and gaining some headaches, like this one, by not using NavigationView. It would have given us this animation for free. 

I can't think of a reason to not use it. Why aren't we?

Pete-Hoch:
Updated the animations for the side view.

No one said we are not using NavigationView. (Although I'm not sure if 'a ton of functionality' is accurate.) But what we are doing is trying to match the design spec. There is time and room for expiration and @bmorris-revacomm is investigating the NavigationView. I have been investigating the non NavigationView way. That's all.

bmorris-revacomm (approved):

jbethanco (approved):
Looks smoother. The disappearance off all sidebar items is a little jarring. Could think about fading and animating the offset instead of the width?

jbethanco:
> Updated the animations for the side view.
> 
> No one said we are not using NavigationView. (Although I'm not sure if 'a ton of functionality' is accurate.) But what we are doing is trying to match the design spec. There is time and room for expiration and @bmorris-revacomm is investigating the NavigationView. I have been investigating the non NavigationView way. That's all.

Still learning the intricacies of your collaboration style. If we are merging to the release branch, I thought that was a decision. If we are exploring various ideas, should we have competing branches that aren't merged?

Pete-Hoch:
> > Updated the animations for the side view.
> > No one said we are not using NavigationView. (Although I'm not sure if 'a ton of functionality' is accurate.) But what we are doing is trying to match the design spec. There is time and room for expiration and @bmorris-revacomm is investigating the NavigationView. I have been investigating the non NavigationView way. That's all.
> 
> Still learning the intricacies of your collaboration style. If we are merging to the release branch, I thought that was a decision. If we are exploring various ideas, should we have competing branches that aren't merged?

So far, using NavigationView has been a non starter. Getting it to cooperate with the current design seem impossible. But we are going to continue to try.

If we had a stable app and we were adding a new standalone feature, it should sit on a feature branch until it's ready to be added to the release. But that is not the case we have right now. Right now, we are building form the ground up. All we have is release/0.2.0.

We could make a staging branch and then land it on release/0.2.0 once everything is done, but I do not see the advantage of doing that for what is essentially a 1.0 release.

I'm open to other suggestions on how to stage our changes. But wherever we make the changes, we should do it so that we can start to combine the data with the views.

PR #145, opened 2021-01-06
Add blank views for Aircrew List & Data and Overview.

Pete-Hoch:
Also plug in a state variable for the current page.

We can now toggle between the Overview, et al from the side menu.


jbethanco (approved):
Runs good.

PR #147, opened 2021-01-06
added views

bmorris-revacomm:
- added some lists (no functionality)
- cleaned some views

The lists look pretty ugly when the menu is open - still working on that part :)


jbethanco (approved):
It runs.

Pete-Hoch (approved):
I keep asking for a spell checker in Xcode, but I never get one.

Looks good.

PR #149, opened 2021-01-07
Start to list the Forms in the sidebar.

Pete-Hoch:
Get previews working again for Forms & Sidebar.
Adjust a couple of colors for dark mode.
Add the form buttons to the side bar. (works in preview)


jbethanco (approved):
Works here. Glad the previews are back!

PR #150, opened 2021-01-07
Fix/enhance persistent data

jbethanco:
While working on the CoreData stack realized we need a way to show system wide modal alerts to beta testers. I'm against using modal popup alerts in general but thought we we probably need it in specific use cases. Originally wrote it with UIKit, but wanted to go pure SwiftUI. 

For the Core Data stack, was wanting a way to ask the user to get us the error data. Hopefully in the future we can do this via Mattermost, Jira, or at a minimum an email. Yet still need a way to call an alert from non-UI code. This gives us that. Want to ask their permission before sending any data out.

Short Version:
Added an alert view provider with a custom view modifier we can attach to the initial content view container.
Added shake to add fake data records during debug. In simulator do Control+Command+Z to add fake records to the persistence. No way to delete yet. 
I think I fixed up the header view a bit in there as well in accordance with the comments. 
Removed the old core data stuff from the app delegate.

FREE steak dinner to the person that:
Figures out a way to prevent a custom view modifier from being added to a view hierarchy more than once. 
e.g. This AlertsProvider modifier can be added more than once to the same view hierarchy. How do we crash that? For some reason, it doesn't cause any problems in this case, but others it could.

### DIFF
--- a/PBLogging.xcodeproj/project.pbxproj
+++ b/PBLogging.xcodeproj/project.pbxproj
@@ -41,6 +41,7 @@
 		7FF0FE7B25A64CC700097C66 /* RegularText.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7FF0FE7A25A64CC700097C66 /* RegularText.swift */; };
 		7FF0FE8325A64D9200097C66 /* ListHeaderText.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7FF0FE8225A64D9200097C66 /* ListHeaderText.swift */; };
 		BF4056F925A555D700872E70 /* Form781+SwiftUIHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF4056F825A555D700872E70 /* Form781+SwiftUIHelpers.swift */; };
+		BF5C1F2425A683690075404C /* AlertProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF5C1F2325A683690075404C /* AlertProvider.swift */; };
 		BF982CC825A54B0500C14019 /* PBLogging.xcdatamodeld in Sources */ = {isa = PBXBuildFile; fileRef = BF982CC625A54B0500C14019 /* PBLogging.xcdatamodeld */; };
 		BF982CD025A54E3D00C14019 /* Persistence.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF982CCF25A54E3D00C14019 /* Persistence.swift */; };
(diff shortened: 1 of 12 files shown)


PBLogging/Views/HeaderView.swift
```diff
@@ -9,25 +9,18 @@ import SwiftUI
...
-
-            // ATTN: Is there a better way to push the
-            // left and right apart?
-            Spacer(minLength: 5)
-
-            // The right side of the bar.
+            Spacer()
             TextAndIconButton(text: "HELP CENTER", color: .white, icon: "info.circle")
         }
-        .frame(height: 50.0)    // ATTN: I don't like hard coding this, what other way can we do it?
```
Pete-Hoch, line 27: I wonder what changed that this can be removed. A few days ago, without it, the header hugged the Image() but now it's given space. I'm glad it's gone, just curious why it works now when it didn't before.

PBLogging.xcodeproj/project.pbxproj
```diff
@@ -716,7 +720,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 0;
 				DEVELOPMENT_ASSET_PATHS = "\"PBLogging/PreviewContent\"";
-				DEVELOPMENT_TEAM = RQ9T47YT3S;
```
Pete-Hoch, line 719: I've seen this change before. Is one team better than the other?

Pete-Hoch (approved):
Not sure how to get the alert to display, but I like the other changes.

My only real question is, why did DEVELOPMENT_TEAM change?

PBLogging.xcodeproj/project.pbxproj
jbethanco: I think it is just my personal business account, so I can test on my associated devices.  Would be useful to get a Glotech or Revacomm account once we get our biz iPads.

PBLogging/Views/HeaderView.swift
jbethanco: Maybe because the Icon image on the left was given a size?

jbethanco:
> Not sure how to get the alert to display, but I like the other changes.
> 
> My only real question is, why did DEVELOPMENT_TEAM change?

To get the an alert to show call:

AlertProvider.shared.showAlertWithTitle(title: "Context Save Error", message: "\(nsError), \(nsError.userInfo). Please screenshot and send to the dev team.")
            
            Or... if you have your own flavor created in the enum:

AlertProvider.shared.showAlert(.yourFlavor)

PR #151, opened 2021-01-07
Hook up HELP CENTER button

Pete-Hoch:
A simple one, but now we can call for help when we need it.


jbethanco (approved):
Works here. 

Just noticed something unrelated:

With a lot of forms... looks like this:

Screen Shot 2021-01-06 at 8 06 10 PM

Pete-Hoch:
> Works here.
> 
> Just noticed something unrelated:
> 
> With a lot of forms... looks like this:

I assume we will need to add some Lists or other view that has scrolling. But for now, let's get the Core Data hooked up to the UI.

PR #152, opened 2021-01-07
Try to smooth out the sidebar animation.

jbethanco:
https://user-images.githubusercontent.com/76016626/103868250-b9461500-506c-11eb-940c-3474ec20d9a9.mp4

Adjust some button sizes.

This isn't a pull demand... just a pull-request. 
Feel free to trash it.
This makes the sidebar animation smooth with an offset move and without making the buttons disappear.
I also added some sidebar button padding to make them not touch the edges of the screen.   Also made them a little shorter.  

Won't merge without full UI team buy-in. Or if the button changes are too much... please snipe the offset code and delete the branch and PR.


PBLogging/Views/SideViews/SideView.swift
```diff
@@ -56,38 +54,41 @@ struct SideView: View {
                             // and use it to set the current Form
                         })
                     }
-                    // Keep this last to push everything up.
```
Pete-Hoch, line 59: Why delete the comment? Is it in error? Is it as obvious as 2 + 2 = 4?

PBLogging/Views/SideViews/SideView.swift
```diff
@@ -56,38 +54,41 @@ struct SideView: View {
...
+                
+                Spacer()
             }
+            
         }
-        .frame(width: isCollapsed ? 40 : 200)
-        .clipped()
+        .frame(width: 240)
+        .offset(x: isCollapsed ? -100 : 0)
+        .frame(width: isCollapsed ? 40 : 240)
```
Pete-Hoch, line 65: I was just thinking earlier that the sideview should be narrower to make more room for the form data.

Why are there two .frame() calls? Maybe some comments would help.

PBLogging/Views/SideViews/SideViewButton.swift
```diff
@@ -20,12 +20,14 @@ struct SideViewButton: View {
         }) {
             HStack {
                 BoldText(text: text, size: 14, color: .pblSlate)
-                    .padding(.all)
+                    .padding(EdgeInsets(top: 10, leading: 8, bottom: 10, trailing: 8))
```
Pete-Hoch, line 23: I like this, but I don't think we need to widen the sidebar because of it.

Pete-Hoch (commented):
Just some comments. I'll play with it in the morning.

PBLogging/Views/SideViews/SideView.swift
jbethanco: It felt obvious. But that's to me. I forget that my instinctual deletions get committed as well.

PBLogging/Views/SideViews/SideView.swift
jbethanco: I'm actually not sure. I was fighting something with the animation at the end when it finally worked. I was probably fighting that second frame call. Theoretically it shouldn't need any frame width animation.

PBLogging/Views/SideViews/SideViewButton.swift
jbethanco: I think I was just fiddling with that. We've got the padding there and outside the button. Try removing the one from the button and see how it looks.

PBLogging/Views/SideViews/SideView.swift
petermisik: I know some people feel comments are clutter, but for those that have left projects for others, they're a source of documentation.   With swiftUI, we know that we need a spacer to push, but is leaving this comment terrible?

petermisik (commented):
Looking good so far.

PBLogging/Views/SideViews/SideView.swift
jbethanco: If we all know the spacer pushes, then leaving a comment that says spacer pushes wouldn't be a good thing. 
I know SwiftUI is new though and some things aren't as obvious to some as they are to others. I'll try to be less generous with the backspace. :)

PBLogging/Views/SideViews/SideView.swift
petermisik: If you drop the top .frame, you'll lose the chevron.

PR #154, opened 2021-01-07
Fix/cleaned UI

bmorris-revacomm:
- header and footer not in list anymore (don't scroll with list)
- header text gets smaller instead of multiple lines
- added some missing buttons


PBLogging/Views/FormViews/AircrewData/FlightTimeListView.swift
```diff
@@ -9,32 +9,35 @@ import SwiftUI
...
     var body: some View {
-        List {
+        VStack {
             FlightTimeListHeader()
-            FlightTimeCellView()
-            FlightTimeCellView()
-            FlightTimeCellView()
+                .padding(.horizontal)
+            ListHeaderLine()
+            List {
```
Pete-Hoch, line 16: Sounds like this can be a ScrollView too. 

Based on what Pete found and if we don't expect to use any List features.

PBLogging/Views/FormViews/FlightSeqListView.swift
```diff
@@ -9,56 +9,48 @@ import SwiftUI
...
-        List {
+        VStack {
             FlightSeqListHeader()
-            FlightSeqCellView()
-            FlightSeqCellView()
-            FlightSeqCellView()
-            AddFlightView()
+                .padding(.horizontal)
+            ListHeaderLine()
+            List {
```
Pete-Hoch, line 16: Again, do we just need a ScrollView?

PBLogging/Views/Misc/ListHeaderText.swift
```diff
@@ -10,19 +10,21 @@ import SwiftUI
 struct ListHeaderText: View {
     let text: String
     let size: CGFloat
+    let lines: Int
```
Pete-Hoch, line 13: Can we give this a default = 1? That is the most common and then we will not have to pass it in to 95% of the usages.

Pete-Hoch (changes requested):
Just a few simple things.

Pete-Hoch (approved):
Looks good.

petermisik (approved):
Looks good to me.  Take a look at Pete's comments though.  It might be easier to drop in the ScrollViews than working with a list, unless there's something I'm missing.

PR #155, opened 2021-01-07
Fix/put days in list

bmorris-revacomm:
- put days into scrollview so they don't get crazy and out of control
- set line limit default to 1


Pete-Hoch (approved):
This looks really good!

1) Mission Data takes a few seconds to load. We should try and figure out why.
2) The form title (with share & print) seems a bit taller than it needs to be.

Both of those things can be worked on as follow up tasks.

PR #157, opened 2021-01-08
Add new form button.

Pete-Hoch:
Tap button, add one new blank for with todays date.


jbethanco (approved):
Does what it says. And says what it does.

PR #159, opened 2021-01-08
Fix/added overview cells

bmorris-revacomm:
- added overview cells
- added default color for Bold and Regular Text
- removed unnecessary parameter values
- updated some Text to match Rachael's updates


Pete-Hoch (approved):
I love it!

PR #160, opened 2021-01-09
Add frames around text fields groups for speed.

Pete-Hoch:
As John found out, we can put multiple TextFields in the same view and still have the view transition fairly snappy if we wrap them in a Form. So this is a patch to try that out. Need Rachel to take a look since this modifies the UI slightly.

There is still a slight delay, so we may have to break each section into a different page. (Or discover something brilliant.)

Screen Shot 2021-01-08 at 4 47 03 PM

Screen Shot 2021-01-08 at 4 47 49 PM

Screen Shot 2021-01-08 at 4 48 16 PM


jbethanco (approved):
It is faster. 
Need more.

PR #161, opened 2021-01-11
Several changes, mostly to the Overview page.

Pete-Hoch:
Display actual form list on Overview page.
Can add or delete or edit forms from Overview page.
Changed Date extension to put out the needed format for a String.


jbethanco (approved):
We definitely need to request using standard DatePickers. Perhaps a meeting with Christian. I can create a quick demo project for download prior so he can test on his iPad.  Can show current way and its issues, as well as the iOS 13 and 14 DatePickers.

PR #164, opened 2021-01-12
place padding/frame on button content, not button itself

bmorris-revacomm:
- fixed refactor from yesterday (put padding/frame on button content instead of button itself so tappable area is larger)


jbethanco (approved):
This was my mistake and incorrect thinking. 

I think we should move foregroundColor and background inside as well. My thinking now is that we should look at the button as the container for the action and the label. The modifiers should be as close to what they are modifying as possible. And they all modify the label and not the action.

jbethanco (approved):
nice.

PR #167, opened 2021-01-13
Bring back PDF printing.

Pete-Hoch:
Also added back the layout we lost for Mission Data.


jbethanco (commented):
I thought we had agreed to do an intermediate preview view prior to the share or print so we would not lock the main thread and could display the Activity Indicator. If implemented as is, and we fill out the form, we are going to have the same lag problem we had before with a frozen UI.

Pete-Hoch:
> I thought we had agreed to do an intermediate preview view prior to the share or print so we would not lock the main thread and could display the Activity Indicator. If implemented as is, and we fill out the form, we are going to have the same lag problem we had before with a frozen UI.

Bringing 0.1.0 functionality over does not prevent us from doing that. But it does give us more ways to test and more things to demo.

jbethanco (approved):
Yea. I'm thinking it would be cool to get Rachael a Github account.

PR #168, opened 2021-01-13
add flight conditions

bmorris-revacomm:
- added flight conditions
- cleaned some code


Pete-Hoch (approved):
We clearly have layout things to clean up, but that is everywhere. 

Looks good all in all.

PR #169, opened 2021-01-13
This is a little clunky, but it gives us the Share sheet.

Pete-Hoch:
Can clean up more later.


Pete-Hoch:
Handled in another branch.

PR #170, opened 2021-01-14
Change Navigation

jbethanco:
Nav view is in.
Reads from core data, but does not write
Lots to do


Pete-Hoch:
As we talked about, let's get a scroller under the form cards on the main page and reduce the (i) to just the current link.

jbethanco:
> As we talked about, let's get a scroller under the form cards on the main page and reduce the (i) to just the current link.

Done.

petermisik (commented):
I like it. I think putting the old files in a folder is good for continuity.

Pete-Hoch (approved):
Let's get going.

PR #171, opened 2021-01-14
Flesh out the mission data.

Pete-Hoch:


jbethanco (approved):
Looks good. 
Getting a trailing ")" in Flight Auth field but I think it's a fake data issue.

PR #172, opened 2021-01-14
Add Complete to the form overview card & align fields.

Pete-Hoch:
Very simplistic layout, but it gets the fields aligned.

We still need a Delete function somewhere.

And a place to put the "REVIEW & SHARE" button, if that still exists in the new design.


jbethanco (approved):
Looks good. 
(We may need to use alignment guides 🤮 to center the icons on the previous rows icons? Still haven't wrapped my head completely around those. )

Pete-Hoch:
> Looks good.
> (We may need to use alignment guides 🤮 to center the icons on the previous rows icons? Still haven't wrapped my head completely around those. )

Agreed, those or the geometry reader and then start each column at 5%, 35%, 60%, whatever. So many ways!

PR #173, opened 2021-01-14
refactor flight seq UI

bmorris-revacomm:
- refactored UI for flight seq (did rounded corners for fun to see what it looked like 😁)


jbethanco (approved):
Looks good to me.
Maybe we can pair code tomorrow and to get the CoreData to show up in more places.

PR #176, opened 2021-01-15
play with saving data changes

Pete-Hoch:
This is a one off of me saving the edit change for MDS in the mission data section. 

I am sure we can find something that is not so clunky, but I wanted to leave the code near the action just so it's obvious.


jbethanco (approved):
These work. The foreground color in darkmode should be the green. I added it as .pblPrimary in the colors extension. Need to update that on the flight cards too. 
I think I saw a CGFloat(of a CGFloat). I'll find it tomorrow.

PR #177, opened 2021-01-15
clean aircrew detail for demo day

bmorris-revacomm:
- cleaned aircrew detail view and corresponding views for demo day (because Pete M will be out that day)
- not all textfields are hooked up correctly - just wanted to get UI cleaned up for the demo

Side note: It works fine if you run it in the ContentView preview pane, but doesn't show up in the simulator... not sure why


jbethanco (approved):
I like it.
Aircrew Detail sections are centered instead of leading. Not sure how to change.

Pete-Hoch (approved):
Since the Aircrew Detail had multiple Sections in a VStack, the VStack was using default alignment .center. Changed to .leading and tightened up the layout.

Please Squash & Merge.

bmorris-revacomm:
@jbethanco @Pete-Hoch Are you both done with my PR yet?? I'm supposed to be sleeping 🙄 😉

PR #179, opened 2021-01-15
clean up previews and add dark mode

bmorris-revacomm:
- made sure all previews have dark mode
- cleaned up some previews so they are more or less formatted the same


Pete-Hoch (approved):
Looks great!

PR #181, opened 2021-01-15
Add de-blur on-appear animation to hide the filling out of the form.

jbethanco:


Pete-Hoch (approved):
Saw it in the demo, looks good.

PR #182, opened 2021-01-15
Make Mission Data persist take 2.

Pete-Hoch:
Work around on(Dis)Appear bug with Sections.

Update the 781 form card on return.

Allow for multiple date entry formats. (Hopefully replaced will a picker later.)

Add relevant unit test back in.

### DIFF
--- a/LoggingTests/LoggingTests.swift
+++ b/LoggingTests/LoggingTests.swift
@@ -9,7 +9,7 @@ import XCTest
 @testable import Logging
 
 class LoggingTests: XCTestCase {
-    
+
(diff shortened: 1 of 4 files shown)


jbethanco (commented):
Just tested it with Group{} ... onAppear and onDisappear still getting called multiple times.
Tried with VStack (which killed the formatting) and they got called less, but still an incorrect amount of times. Think we should explore this problem further. 

Tested the lifecycle functions off the main form VStack. They work correctly. Think we should change the section States to bindings and put the states in the main form.

PR #183, opened 2021-01-18
Remove unused PDF generator file

jbethanco:


petermisik (commented):
Works.

PBLogging/OldNavViews/FormViews/FormView.swift
```diff
@@ -37,9 +37,9 @@ struct FormView: View {
 //                        }
                     }
                     TextAndIconButton(text: "PRINT", icon: "printer.fill") {
-                        if let form781 = forms.first {
-                            form781.printPDF()
-                        }
+//                        if let form781 = forms.first {
+//                            form781.printPDF()
+//                        }
                     }
```
Pete-Hoch, line 43: Since the code has been removed, we may as well remove this too. Since the name printPDF() was reused, this will show up on searches through the code and could be confusing. In fact, since PersistenceController.newRecordForContext() is being called elsewhere, I think it's reasonable to remove this whole file form the project now.

Pete-Hoch (changes requested):
Let's get rid of FormView.swift too.

Pete-Hoch (approved):

PR #184, opened 2021-01-18
Make single convenient method for saving Managed Object Context on Pe…

jbethanco:
…rsistence. Call with "Persistence.saveContext()"

Replace current call sites for the above.
Rename FauxData to FakeData as Faux is a less common word.
Move fake data functions that were in Persistence to FakeData so responsibility is more clear.
Add Convenient functions to FakeData that generate fake Flight and fake AircrewData arrays for use in SwiftUI Canvas Previews.
Update all previews to use new FakeData Flight and AircrewData methods


petermisik (commented):
This looks good.

Pete-Hoch (approved):
Good clean up.

PR #187, opened 2021-01-19
refactor form cards

bmorris-revacomm:
- refactored form cards based on Rachael's new designs


PBLogging/NewNavViews/ContentHeaderView.swift
```diff
@@ -19,8 +19,12 @@ struct ContentHeaderView: View {
                     UIApplication.shared.open(url)
                 }
             } label: {
-                Image(systemName: "info.circle")
-                    .padding()
+                HStack {
+                    Text("Help Center")
+                        .font(.pblBold(size: 18))
+                    Image(systemName: "info.circle")
```
Pete-Hoch, line 25: In Figma the icon is at least as big as the text. So let's start by making the image size the same as the font size.

PBLogging/NewNavViews/ContentHeaderView.swift
```diff
@@ -19,8 +19,12 @@ struct ContentHeaderView: View {
                     UIApplication.shared.open(url)
                 }
             } label: {
-                Image(systemName: "info.circle")
-                    .padding()
+                HStack {
```
Pete-Hoch, line 22: We should add the PBL image on the left also.

PBLogging/NewNavViews/ContentHeaderView.swift
```diff
@@ -33,6 +37,8 @@ struct ContentHeaderView: View {
 //            }
             
         }
+        .background(Color.pblNotBlack)
```
Pete-Hoch, line 40: The background color does not match Figma, but let's take the color assignment as a different task.

PBLogging/NewNavViews/Form781Card.swift
```diff
@@ -16,62 +16,36 @@ struct Form781Card: View {
...
-                }
-                .padding(1)
+                Text("AFTO FORM 781")
+                    .font(.pblBold(size: 18))
+                IconText(image: "calendar", text: "21 Jul 2020")
+            }
+            Spacer()
+            VStack(alignment: .leading) {
+                IconText(image: "square.and.arrow.up", text: "Last shared:")
+                IconText(image: "envelope", text: "Last emailed:")
```
Pete-Hoch, line 28: I believe we will need two layout definitions one for portrait and one for landscape.

Let's take that as a different task.

Pete-Hoch (changes requested):
A few things to fix and a few things to take as new tasks.

Pete-Hoch (approved):
Looks good.

PR #188, opened 2021-01-19
Add PreviewMods file with:

jbethanco:
-Easy to use preview modifiers providing device, title, and dark mode in one line.
-Provides Mock wrapper for dynamic values in previews.
-Change two call sites to demonstrate usage.

Just creature features for devs. In PreviewContent folder so doesn't ship.


petermisik (approved):
Works great.

PR #190, opened 2021-01-19
Add new figma design colors

Pete-Hoch:
Bring over the design colors from Figma and add then as their new names.

Did not switch over to them other than making Primary our accent color.


petermisik (commented):
I like it.  The IconText idea is really cool.

jbethanco (approved):
Works good. Does the IconText lock in our alignment problems though?

Pete-Hoch:
IconText was not actually an addition in this PR and I think Beth already removed it due to the alignment issues.

bmorris-revacomm:
I didn't remove the file, but I did change those two fields because they were the only ones affected. Do you think I should remove the file?

Pete-Hoch:
> I didn't remove the file, but I did change those two fields because they were the only ones affected. Do you think I should remove the file?

I don't think that is necessary. It's a nice item where we can use it.

PR #193, opened 2021-01-20
Add landscape layout for the Overview window.

Pete-Hoch:
This is still based on the device orientation and the full screen size and not on the app window size. But I'm not sure if we care about the multi app case, if we go there we may as well be doing a phone app also.


jbethanco (approved):
Works good on device.

PR #195, opened 2021-01-20
add font modifiers

bmorris-revacomm:
- added font modifiers file from what Rachael has for the Typography


jbethanco (approved):
Nice. Previews look awesome.

PR #196, opened 2021-01-20
Fix/enhance animation

jbethanco:
Played around with this last night:
- Labels will animate movement to new positions on devices with iOS 14 and fade in out as they do currently on iOS 13
-Make CardLabel, for card view, to simplify animation, spacing, and alignment.
-Add WrappedNamespace file to wrap namespace call for geometry effect
-Change Landscape/Portrait calls to remove optional, and avoid unnecessary or repeated animations for non Landscape or Portrait orientations. (e.g. faceDown, faceUp, and unknown)


PBLogging/NewNavViews/Form781View/Overview/Form781Card.swift
```diff
@@ -92,3 +92,25 @@ struct Form781Card_Previews: PreviewProvider {
...
+    let mutableText: String
+    var imagePadding: CGFloat = 0.0
+    
+    var body: some View {
+        HStack {
+            Image(systemName: image)
+                .padding(.horizontal, imagePadding)
+            Text(labelText)
+                .font(.pblRegular(size: 16))
+            Spacer()
```
Pete-Hoch, line 110: I don't think we want a spacer here. There is a gap in the Figma, but the editable text is much closer to the associated label and this spaces out equal from the two labels in landscape.

Pete-Hoch (changes requested):
Looks nice, but landscapes layout does not match Figma.

jbethanco:
I think the fix is adjusting the card label frame width. I left it wider
than needed as wasn’t sure we were done playing with the fonts.

jbethanco:
> Looks nice, but landscapes layout does not match Figma.

Think the width change fixed it.

jbethanco (commented):
fixed

Pete-Hoch (approved):
Nice.

PR #197, opened 2021-01-20
updated fonts in form view

bmorris-revacomm:
- updated fonts in the form view
- also updated font sizes for font modifiers because I'm a dum-dum and had put the height instead of font size :)


Pete-Hoch (approved):
Looks good.

PR #201, opened 2021-01-21
add section header modifier

bmorris-revacomm:
- added section header modifier (allows lowercase text, but only available in iOS 14)
- fixed some UI colors and spacing
- removed help center text


Pete-Hoch (approved):
This looks good.

PR #202, opened 2021-01-21
add new flight seq view

bmorris-revacomm:
- added new flight seq view (kept old one as well, because i'm not sure if the new one is more of a detail view or not?) (old cards navigate to new view)
- added "inactive" color to color extension


Pete-Hoch:
Screen Shot 2021-01-21 at 11 30 18 AM

I think we need to fix this. When I click to go to the next page, I get this highlight and it does not go away until I click on another item.

PR #203, opened 2021-01-22
Move header view out of navigation bar items.

jbethanco:
NavigationLinks don't seem to work in navigationBarItems, so the PDF Preview button wasn't firing.
This puts it back how it was. We'll need to figure out another way.


jbethanco:
Tried to move it up to the desired position using: ".padding(.top, -100)"
It moved it up, but it wasn't tappable.

Pete-Hoch (approved):
Looks good.

PR #204, opened 2021-01-22
Let the old pass away.

Pete-Hoch:
Checkout hash: 392a4fa45e79f5f33b97f28ff6cee340de23ebb9 to see it again.

We now have more functionality in the new UI code than the old. So I think it's time to remove the old code. But since git never forgets, feel free to checkout a branch at the hash noted above if you want to go back and revisit that old friend.


jbethanco (approved):
yes!

PR #207, opened 2021-01-22
Capture sharing and printing times

jbethanco:
Replace ActivityViewController (poorly named) with ShareView that has
a callback which we can use to capture the last time the form was shared
or printed.

Capture printed date in printPDF function.


PBLogging/NewNavViews/PDFPreview/PDFPreviewView.swift
```diff
@@ -74,7 +78,23 @@ struct PDFPreviewView: View {
...
+        if let error = error {
+            // TODO: Figure out what kind of errors can happen here and handle approprately vs this:
+            AlertProvider.shared.showAlertWithTitle(title: "Share error", message: "\(error), \(error.localizedDescription). Please screenshot and send to the dev team.")
+            return
+        }
+        if completed {
+            if activityType == UIActivity.ActivityType.print{
+                form.lastPrinted = Date()
+                NSLog("Form Last Printed: \(String(describing: form.lastPrinted))")
+            }else{
```
Pete-Hoch, line 92: Can we make this an explicit list of activities and then log a message when there is a non sharing activity.

For example is saving the PDF to local files sharing? It just feels like we can be listing things as shared that were not.

Pete-Hoch (changes requested):
Can we also get the UIActivity Category? Then we can record a share for Category share and not action.

PBLogging/NewNavViews/PDFPreview/PDFPreviewView.swift
jbethanco: We need to have a discussion on what is considered a share with Christian.
This PR gets the initial functionality in place. We can get more granular after feedback.

Pete-Hoch: The OS already makes a distinction between UIActivity.Category.share and UIActivity.Category.action. Can we use this?

Pete-Hoch (approved):
OK for now.

PR #208, opened 2021-01-25
Add initial Core Data state for UI tests.

jbethanco:
Delete all 781s on app launch for UI Tests
Generate standard fake set of Form 781s for UI Tests
Add clearCoreData function to Fake Data for easy clearing of all 781s


LoggingUITests/LoggingUITests.swift
```diff
@@ -25,6 +25,7 @@ class LoggingUITests: XCTestCase {
     func testExample() throws {
         // UI tests must launch the application that they test.
         let app = XCUIApplication()
+        app.launchArguments.append("CLEAR_CORE_DATA_THEN_LOAD_FAKE_DATA")
         app.launch()
```
Pete-Hoch, line 29: Should we do this in setup() so that every test will get this?
And if we do setup() then how about the "CLEAR_CORE_DATA" in teardown()?
That way, every test gets the same staring point.

jbethanco: I think that should be up to the preference of the UI Test writer.  Can't tear it down, via the current method as it occurs during app launch. Would have to write a separate function for that. Not sure if globally clearing on app close would be appropriate as  as it could cause some unintended consequences when we test backgrounding and return states. Keeping it per tests allows for the most granular control and clearest intent.

Pete-Hoch: Agreed, a separate function is a good idea. Let's add this as a test utility so the test written can just call something like PBLTestUtil.clearAllData or PBLTestUtil.setupFakeData.

petermisik: I like the PR.  I think there is something to be said for not leaving data behind after testing, but that should be at the discretion of the test writer.  I think they should have it cleared on the final test as to not leave any lingering data.

jbethanco: > Agreed, a separate function is a good idea. Let's add this as a test utility so the test written can just call something like PBLTestUtil.clearAllData or PBLTestUtil.setupFakeData.

I think you may be misunderstanding on how UI Test works. We can't just call it like commands. We can  only append a launch argument flags and have it do things based on a passive observation of those flags. Appending flags at the beginning of tests makes the intent clear and works. We could add a "clear core data on close" flag, but that obscures the intent from the next test that should just append the flag to clear on launch. Each individual test should take care of its own work. Does that make sense?

Pete-Hoch (approved):
I'll fix it when I write some UI tests.

PR #209, opened 2021-01-25
PBS-278: Conversion from whole minutes to tenths of an hour 781 logic.

Pete-Hoch:
This is the conversion logic from the 781 form to turn minutes into tenths of an hour.

I change the input and output to numbers (instead of strings) because that seemed more flexible. 

I did not hook this up to be used anywhere, but I think that is OK as another task.


jbethanco:
Would this be better to keep where it is at and bring it over if/when we need to use it?

Pete-Hoch:
> Would this be better to keep where it is at and bring it over if/when we need to use it?

I think it's when we need it. Is there an issue of bringing it over now? So it's there when we need it.

petermisik:
I vote for bringing it.

jbethanco:
> > Would this be better to keep where it is at and bring it over if/when we need to use it?
> 
> I think it's when we need it. Is there an issue of bringing it over now? So it's there when we need it.

No huge deal. It's just odd practice to bring over unused code and tests on unused code.  I think it would be more proper to bring it over when and if it is needed.  Just seems like something that should be done in place upon implementation.

jbethanco (approved):

petermisik (commented):
Looks good.

PR #210, opened 2021-01-26
Add unit tests for String+Ext.

Pete-Hoch:
Gets us over 20%


jbethanco (approved):
It works nice.
(Prefer to delete unused code, rather than comment it out. We can always find it if needed.)

Pete-Hoch:
> (Prefer to delete unused code, rather than comment it out. We can always find it if needed.)

I agree if I wrote the code, or it clearly is not needed. But with everything new, I error on caution.

Added: https://jira.il2.dso.mil/browse/PBS-487

PR #212, opened 2021-01-27
add navigation ui test - PBS-282

bmorris-revacomm:
- added ui test for navigation (will need to update when we add more views)
- updated ui test for adding forms
- changed print button text to "Preview"


PBLogging/NewNavViews/Form781View/FlightsSection/SingleFlightRowView.swift
```diff
@@ -13,6 +13,7 @@ struct SingleFlightRowView: View {
     
     var width:  CGFloat = 100
     var height: CGFloat = 40
+    
     var body: some View {
         VStack (alignment: .leading){
             HStack {
```
Pete-Hoch, line 19: I think you can revert this change and thus the whole file.

Pete-Hoch (approved):
This takes test coverage from 20% to almost 50%, fantastic!

Change in SingleFlightRowView.swift not needed.
Please reference PBS-282 in the title.

Fix, then squash & merge.

jbethanco (approved):

PR #213, opened 2021-01-28
Close PBS-491 - Implement Share Date Capture

jbethanco:
Add share date capture on share completion for email or Mattermost shares only. (Per Leah interview)
Restrict as many off-device share types and unreasonable share types as possible.


Pete-Hoch (approved):
Looks good.

PR #215, opened 2021-01-29
Fix/refactor aircrew list

bmorris-revacomm:
- refactored aircrew list view (kept old ones just in case)


Pete-Hoch (approved):
Looks good.

PR #216, opened 2021-02-03
Delete unused commented out code from String Extensions

jbethanco:
Not a string extension
Unused
We can use string interpolation more safely if something like this is needed.

Closes PBS-487


Pete-Hoch (approved):
Looks good.

PR #217, opened 2021-02-04
Close PBS-523 Display Version/Build number next to logo

jbethanco:
Remove action code out of view code
Remove unused InfoModalView
Remove InfoModalView call site and state var


petermisik (approved):
I like it. The lettering is subtle.

PR #219, opened 2021-02-09
Feature/web view

petermisik:
Here's the code for webview and the scrollview that worked on.  It's not hooked to anything at the moment.  It's here for review.


PBLogging/ScrollView/Cell_label.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  Cell_label.swift
+//  scrolling
+//
+//  Created by Pete Misik on 2/3/21.
+//
+
+import SwiftUI
+
+struct Cell_label: View {
```
jbethanco, line 10: Swift struct names should be camel case. If this is specific to a single use case, this name might be too generic.

jbethanco (commented):
A lot of good here. 

If it's not hooked up or used, we shouldn't merge it though. Fine to keep in another branch for storage until it is used. Ideally we should have no unused code in the release branch. 
YAGNI

PBLogging/WebView/Loader.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import SwiftUI
+
+struct Loader: View {
+    @State var spinCircle = false
+    
+    var body: some View {
+        ZStack {
+            Rectangle().frame(width:160, height: 135).background(Color.black).cornerRadius(8).opacity(0.6).shadow(color: .black, radius: 16)
```
Pete-Hoch, line 15: Please put each .item on a new line.

Pete-Hoch (commented):
Good stuff here! We should be able to start hooking it up soon.

PR #220, opened 2021-02-11
Add the new Events overview page.

Pete-Hoch:
This is just the new Event List page as seen in Figma.

The code is still using the current Core Data Form781, but it should be simple to swap it out when the new data model is added.

I'm using harmLocation for the eventName just to have some data.


jbethanco (dismissed):
The modal isn't behaving modally. You can bring up more than one and also click behind. See attached video for visual.

The call site for the modal is kind of confusing. I was working on something similar the other day for the same task, that is called similarly to SwiftUI's sheet.  It's pretty reusable as well, so we could use it for other modal popups. You can see it here: 

Feel free to steal. 
https://github.com/jbethanco/DataModel/blob/master/DataModel/Views/PBLModal.swift

It currently switches off the bound bool with the completion handler, but it might be cleaner to make it a binding on the modal itself.
 
https://user-images.githubusercontent.com/76016626/107601411-be960280-6bca-11eb-8755-2c428140e831.mp4

Pete-Hoch:
> The modal isn't behaving modally. You can bring up more than one and also click behind. See attached video for visual.
> 
> The call site for the modal is kind of confusing. I was working on something similar the other day for the same task, that is called similarly to SwiftUI's sheet. It's pretty reusable as well, so we could use it for other modal popups. You can see it here:
> 
> Feel free to steal.
> https://github.com/jbethanco/DataModel/blob/master/DataModel/Views/PBLModal.swift
> 
> It currently switches off the bound bool with the completion handler, but it might be cleaner to make it a binding on the modal itself.
> 
>  modality.mp4

Added this task: https://jira.il2.dso.mil/browse/PBS-583

So when we have a design, we can do the modal dialog as specified. In the mean time, this is just a wireframe modal dialog and not really the major point of this PR.

Can we land the view so that we can get the rest of the views connected to it?

bmorris-revacomm (approved):
looks good!

jbethanco:
> > The modal isn't behaving modally. You can bring up more than one and also click behind. See attached video for visual.
> > The call site for the modal is kind of confusing. I was working on something similar the other day for the same task, that is called similarly to SwiftUI's sheet. It's pretty reusable as well, so we could use it for other modal popups. You can see it here:
> > Feel free to steal.
> > https://github.com/jbethanco/DataModel/blob/master/DataModel/Views/PBLModal.swift
> > It currently switches off the bound bool with the completion handler, but it might be cleaner to make it a binding on the modal itself.
> > modality.mp4
> 
> Added this task: https://jira.il2.dso.mil/browse/PBS-583
> 
> So when we have a design, we can do the modal dialog as specified. In the mean time, this is just a wireframe modal dialog and not really the major point of this PR.
> 
> Can we land the view so that we can get the rest of the views connected to it?

Let's delete the modal code and then merge so we don't have code we know we aren't going to use in the base.

Pete-Hoch:
> > > The modal isn't behaving modally. You can bring up more than one and also click behind. See attached video for visual.
> > > The call site for the modal is kind of confusing. I was working on something similar the other day for the same task, that is called similarly to SwiftUI's sheet. It's pretty reusable as well, so we could use it for other modal popups. You can see it here:
> > > Feel free to steal.
> > > https://github.com/jbethanco/DataModel/blob/master/DataModel/Views/PBLModal.swift
> > > It currently switches off the bound bool with the completion handler, but it might be cleaner to make it a binding on the modal itself.
> > > modality.mp4
> > 
> > 
> > Added this task: https://jira.il2.dso.mil/browse/PBS-583
> > So when we have a design, we can do the modal dialog as specified. In the mean time, this is just a wireframe modal dialog and not really the major point of this PR.
> > Can we land the view so that we can get the rest of the views connected to it?
> 
> Let's delete the modal code and then merge so we don't have code we know we aren't going to use in the base.

I'd prefer to have a way to add something new. I chan change it to red with a "DEVELOPMENT ONLY" Title if that will help. Call me silly, but I like the idea of adding random text instead of just loading pre-baked data by shaking my iPad. I'd prefer to leave it in for now. It's the best way for Rachael to play with it and see something there too.

petermisik:
I think the "DEVELOPMENT ONLY" tag is a good compromise.  Let's leave it for now and re-attack when we have a solid design.

petermisik (approved):
See my other comment.  Good compromise.

Pete-Hoch:
I believe the last push addresses all requested changes.

PR #221, opened 2021-02-11
Fix/update mission data view

bmorris-revacomm:
- updated mission data view based on Rachael's changes. Left the old views, just in case.
- still needs some UI tweaks and functionality
- added tabs (only the first tab item is linked correctly - waiting on other views)
- removed old UI tests

- I think I accidentally branched this off of Pete H's branch, so some of his changes are in here as well. Is this a problem?
- (I reverted my change out of your branch, so that. is gone now.)


jbethanco (approved):
It works. It tabs. It breaks UI tests. :) 

Will need some updates for Dark Mode.

Pete-Hoch (changes requested):
Let's get the columns sized correctly for the "COMBAT TIME SRTY" & "COMBAT SUPPORT Time SRTY" columns. (It may just be a header size issue.)

Pete-Hoch (approved):
New changes look good.

PR #222, opened 2021-02-11
add "elevated" color

bmorris-revacomm:
- added "elevated" color
- updated column color
- organize folders/files


Pete-Hoch (approved):
Looks good.

PR #224, opened 2021-02-12
Fix/add sortie comments tab

bmorris-revacomm:
- added sortie comments tab
- added UIColor extension for textview


Pete-Hoch (approved):
Looks good.

PR #228, opened 2021-02-16
add 781 remarks to mission data view

bmorris-revacomm:
- added remarks section to mission data view
- added "default" to UIColors
- updated text view background color for sortie comments


Pete-Hoch (approved):
Looks good.

PR #230, opened 2021-02-16
Remove dead code and restructure View file groups.

Pete-Hoch:
This removes old code that is no longer used and updates the View group structure to hopefully be easier to navigate in the project.


jbethanco (approved):
Nice.

PR #233, opened 2021-02-18
PBS-281 and 507 - Implement New Data Model

jbethanco:
Close PBS-281 Implement Core Data in .02
Close PBS-507 Implement New Data Model
Close PBS-456 Save Core Data after Page Change

- Replace core 781 Data model with more robust Event data model
- Replace the PersistenceController Singleton with
a non-singleton DataController for easier testing
- Replace MockData with SampleData
- Add Core Data Tests
- Repair connections from views to data models
- Change comments TextView to bindable version
- Fix all Previews for core data and sample data changes


PBLogging/Data/DataController.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        if inMemory {
+            container.persistentStoreDescriptions.first!.url = URL(fileURLWithPath: "/dev/null")
+        }
+
+        container.loadPersistentStores { _, error in
+
+            if let error = error {
+                fatalError("Fatal error loading store: \(error.localizedDescription)")
+               // AlertProvider.shared.showAlertWithTitle(title: "Load Persistent Store Error", message: "\(error), \(error.userInfo). Please screenshot and send to the dev team.")
```
Pete-Hoch, line 23: Please delete comment.

PBLogging/Data/DataController.swift
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    func save() {
+        guard container.viewContext.hasChanges else { return }
+
+        do {
+            try container.viewContext.save()
+        } catch {
+            // Replace this implementation with code to handle the error appropriately.
```
Pete-Hoch, line 35: Please add a task in Jira and reference it here so we do not forget.

LoggingTests/PBLMathTests.swift
```diff
@@ -1,53 +0,0 @@
...
-
-//    override func setUpWithError() throws {
-//        // Put setup code here. This method is called before the invocation of each test method in the class.
-//    }
-//
-//    override func tearDownWithError() throws {
-//        // Put teardown code here. This method is called after the invocation of each test method in the class.
-//    }
-
-    func testDecimalTime() {
```
Pete-Hoch, line 21: Instead of deleting this, can we test tenthsOfAnHour() (or even better, timeBetweenDatesFor781 ()) with the same code coverage range as this test?

Pete-Hoch (approved):
Looks really good. Just a few minor requests.

PBLogging/Data/DataController.swift
jbethanco: Had meant to remove the fatal and add the alert. Will correct. Thanks.

PBLogging/Data/DataController.swift
jbethanco: Will replace with the desired alert provider.

LoggingTests/PBLMathTests.swift
jbethanco: tenthsOfAnHour() is private.
timeBetweenDatesFor781 is private.

calculatedTotalFlightTimeFor781 tests both.
image

jbethanco: timeBetweenDatesFor781() and tenthsOfAnHour() are private.
Both are covered with the testSortieTotalTime() test. It calls "sortie.calculatedTotalFlightTimeFor781" which uses both private methods. 

image

Pete-Hoch: Ah cool, I didn't see the upper test.

Pete-Hoch (approved):
Looks good. Let's take it for a walk.

PR #234, opened 2021-02-19
PBS-629 Expand UI Unit Test coverage

Pete-Hoch:
I noticed that out Unit Test code coverage was under 20%, mostly due to the UI tests being commented out.

This restores the UI Unit Tests and conformed to the new views.


PBLogging/Views/TabViews/MissionDataTab/AircrewListView.swift
```diff
@@ -25,7 +25,8 @@ struct AircrewListView: View {
                     Image(systemName: "plus.circle")
                 }
                 .fontSectionHeading()
-            }
+                .accessibility(identifier: "addAircrewButton")
```
jbethanco, line 28: One thing to think about:
While useing accessibility identifiers make it easy to find elements to test, they do clutter the view's code.

Buttons can be called by their labels, such as 
"self.app.buttons["Add Aircrew"].tap()"

This prevents the need to add an identifier modifier to the view. 

In order to find what UI elements are called, you can set a breakpoint in the test code, and then call "po app" in lldb. Or if searching for buttons specifically "po app.buttons"

There is a demo in this video around the 17 minute mark:
https://developer.apple.com/wwdc19/413

jbethanco (approved):
Works good here.

PBLogging/Views/TabViews/MissionDataTab/AircrewListView.swift
Pete-Hoch: Agreed, it works both ways.

PR #235, opened 2021-02-19
PBS-621 and 513 - Allow user to give feedback

jbethanco:
- Add a "CollectorWebView" which loads the destination page of a JIRA feedback collector.
- Inject javascript into webpage to add the build number and version so the user does
not have to manually enter it.
- Add action sheet off of the info button with two options:
1. View Release Notes
2. Submit a Problem


Pete-Hoch:
Screen Shot 2021-02-19 at 7 41 18 AM

The window does not automatically close. (At least it has not after a minute.)

jbethanco:
> Screen Shot 2021-02-19 at 7 41 18 AM
> 
> The window does not automatically close. (At least it has not after a minute.)

We have no control over that part of the Jira collector. 
Do you have a suggestion on way to fix without a Jira callback?

Pete-Hoch:
That is a sadness, we can take it as a bug and deal later. Have added this to Jira.

https://jira.il2.dso.mil/browse/PBS-635

Pete-Hoch (approved):
Cool.

petermisik (approved):
I like the idea of the invisible button.  Good work around.  As always, yay for more WebKit.

jbethanco:
> That is a sadness, we can take it as a bug and deal later. Have added this to Jira.
> 
> https://jira.il2.dso.mil/browse/PBS-635

Thanks.

PR #236, opened 2021-02-19
Fix app start flags for UI tests.

jbethanco:
Flags were not working after to change to non-singleton data controller.


Pete-Hoch (approved):
Looks good.

Too bad we have to be destructive. It would be nice to restore the original data after the test. But that is the life of a developer.

PR #238, opened 2021-02-19
Hook up data fields for Flight SEQ

Pete-Hoch:
This hooks up everything in Flight Seq other than takeoff and landing times. (This are switching to dates.)

Also added a number to a new event (because I lose track when they are all the same name.)


Pete-Hoch:
This was done in another PR.

PR #239, opened 2021-02-20
connect sortie data to mission sortie info view

Pete-Hoch:
Start to hook up the edit fields to the Sortie data.

Note: The Sortie object did not have a fuel, sortieType, or metrics object, so I am adding them, but it's probably not right.

Also, some changes are not sticking. (But all of the toggle are.) So input on that would be appreciated.


PBLogging/Views/TabViews/MissionSortieInfoTab/MissionSortieInfoView.swift
```diff
@@ -8,28 +8,51 @@
...
+    @ObservedObject var sortie: Sortie
+
     var body: some View {
-        ScrollView {
+        // TODO: These items are not being added when we make a new Sortie object.
+        // They are also not in the Sample data. This is probably not the best
+        // place to do this, but I am not sure where is.
+        if sortie.fuel == nil {
+            sortie.fuel = Fuel(context: viewContext)
+        }
```
jbethanco, line 20: I don't know why, but I thought I had done these in the Sortie accessors. I will refactor after this is merged.

jbethanco (approved):
Not sure about the saving thing. I'll explore more.

PR #240, opened 2021-02-21
Fix/add more entity accessors

jbethanco:
Add Sortie Accessors and Hookup and validate some Fuel numbers.
- Add metric, sortieType, and fuel accessors to Sortie entity so they are created upon first access.
- Change all float types in data model to doubles. (precision errors were immediate with floats)
- Correct spelling error of "auxiliaryPowerUnitHours"
- Add "enforceDecimalNumber" extension on string with supporting functions
- Remove one of the unnecessary geometry readers (Per apple techs, should only use if absolutely necessary and most times it isn't necessary. When using to set other sizes it can break expected behavior of the auto layout system.)
- SortieFuelEditFields:
-- Add initializer to prevent having to pass in child property and to set State vars.
-- Use onChange extension on bound State vars to handle the setting of data, and validation of text.
(using NumberFormatter with TextField seemed to not pass data back associated to ObservedObject. It seems to be a bug. Fails with State vars as well. UPDATE: NumberFormatter with textfield does work, but only if the user presses return and if the number formatter doesn't fail, which we have no way of knowing. )


Pete-Hoch (approved):
Looks good, thanks!

PR #241, opened 2021-02-22
PBS-642 Fix Core Data Errors during unit tests

jbethanco:
Core Data does not cache the data model. During unit tests we bring up
 a second data controller for testing. This causes Core Data to load the
data model twice. It then sees conflicts between the two (same)
data models. By making the model a static let off of the dataController
class it can not be loaded twice.

Minor unrelated change:
Wrap  checkForUITestArguments() with Debug flag so it doesn't go into
production code.


Pete-Hoch:
This looks fine, but I do not see the bug. How do I get the code to fail? All of our unit tests run fine with or without this patch.

PBLogging/App/SceneDelegate.swift
```diff
@@ -21,7 +21,9 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     
     func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
 
+        #if DEBUG
         checkForUITestArguments()
+        #endif
```
Pete-Hoch, line 26: Good idea.

Pete-Hoch (approved):
Looks good.

petermisik (approved):
Looks good.

PR #242, opened 2021-02-22
Speed up UI testing

jbethanco:
Minor change that disables animations during UI testing by calling:
UIView.setAnimationsEnabled(false)

This speeds up UI testing as we don't have to wait on the animations to happen before querying UI elements.
Also cleaned up and lowercased the testing flags.


Pete-Hoch (approved):
Nice little speed bump.

PR #244, opened 2021-02-23
UI Refactoring

jbethanco:
Removed some Horizontal scroll views. *
Modify chip design *
Added Padding to get some elements off the edges of the screen
Modified some elements to look closer to design
Removed some geometry readers.
Made title of previous "Recent Sorties" page to be dynamic based on event *
* Rachael approved


PBLogging/Views/RecentSorties/SortieCard.swift
```diff
@@ -55,16 +39,16 @@ private struct BodyContent: View {
...
             HStack(alignment: .top, spacing: 30) {
                 ContentColumn(imageName: "doc.on.doc", title: "Sortie Summary", leftColumnLabels: ["Serial Number:", "Mission Number:", "Mission Symbol:", "Special Use:"], rightColumnLabels: [sortie.serialNumber, sortie.missionNumber, sortie.missionSymbol, sortie.specialUse])
-                    .frame(width: viewWidth * 0.32)
+                    .frame(width: 154)
                 ContentColumn(imageName: "location", title: "Location", leftColumnLabels: ["From:", "To:"], rightColumnLabels: [sortie.takeoffICAO, sortie.landICAO])
-                    .frame(width: viewWidth * 0.15)
+                    .frame(width: 90)
                 ContentColumn(imageName: "clock", title: "Flight Time", leftColumnLabels: ["Take off time:", "Land time:", "Total time:"], rightColumnLabels: [sortie.takeoffTime?.string() ?? "", sortie.landTime?.string() ?? "" , sortie.flightTimeString])
-                    .frame(width: viewWidth * 0.2)
+                    .frame(width: 200)
```
Pete-Hoch, line 51: Doing this prevents the columns from spreading out when in landscape. Is that really what we want?

Pete-Hoch (commented):
Looks nice, just one question.

PBLogging/Views/RecentSorties/SortieCard.swift
jbethanco: That's something we could consult with Rachael on. Personally I think it is more visually appealing to not have white space between the columns of text I'm reading.

Pete-Hoch: OK, well then perhaps we should center the columns, or at least limit the last column, because it looks really odd to have the divider line stretch all the way to the trailing edge of the cell, but the text is still bunched up on the leading edge.

PR #248, opened 2021-02-23
Fix/refactor UI

jbethanco:
Refactor Sortie Card
Refactor Loadmaster Stepper
Modify Choice Chips
Add padding in multiple places
Remove horizontal scrollers


Pete-Hoch (approved):
Looks crisp.

PR #249, opened 2021-02-23
update colors

bmorris-revacomm:
- update colors based on design library
- add background color for textview
- remove border for 781 remarks textview


Pete-Hoch (approved):
Good cleanup.

PR #250, opened 2021-02-23
update colors

bmorris-revacomm:
Re-do...
- update colors to match design library
- update textView background color
- remove 781remarks border


Pete-Hoch (approved):
Still looks good.

jbethanco:
Missed a color change so it doesn't run. It's in BoldText() I'll push a fix.

PR #252, opened 2021-02-24
Add unit test for utility extensions

Pete-Hoch:
This adds unit tests for:

Event+Calculations.swift
SortieType+Accessors.swift

Bringing their coverage from 0% to 100%

The test does show an inconsistency for allSortiesCalculatedTime, if there are no Sorties in the event the call returns "0.0". But if there is a Sortie, but either the takeoff time or land time is nil, it returns "". I image we want one or the other of these. But since we do not display anything if there is no Sortie in the Event, we will effectively never return the "0.0".


jbethanco (approved):
This is testing a function off of Event should it be in the Events Test file?

We should probably relook at allSortiesCalculatedTime.
It doesn't return anything if they haven't landed from a third sortie. Perhaps we should report event in progress duration as well. I think that would be at least something interesting to look at on a long TDY. We should probably also test to make sure the math for multiple sorties is happening correctly.

Pete-Hoch:
> This is testing a function off of Event should it be in the Events Test file?
> 
> We should probably relook at allSortiesCalculatedTime.
> It doesn't return anything if they haven't landed from a third sortie. Perhaps we should report event in progress duration as well. I think that would be at least something interesting to look at on a long TDY. We should probably also test to make sure the math for multiple sorties is happening correctly.

I started with the tests in Events, but since they were testing Sortie things, it seemed to make sense to put it here, along with the sotrieType test. But it's fair to move it to Events.

I write up this so we remember to add multiple Sorties and we can move the test function then.

https://jira.il2.dso.mil/browse/PBS-662

PR #253, opened 2021-02-24
Fix/update data model for mission events

jbethanco:
This starting point for the mission event records view. 

Disconnect Mission Event Types from the data model relationships.
Add accessors to MissionEventTypes
Delete/Generate Sample Mission Event Types on launch for development
Starts a Mission Events Records View. It updates and saves, but looks and operates nothing like design. 
Adds "isSim" to Event entity. 
Will probably denormalize the MissionEventTypes further to make it easier to work with rather than a better database design.


Pete-Hoch (approved):
Looks good, obviously the Training Events page needs work.
But that is not the focus of this PR. 
Data changes seem reasonable.

PR #254, opened 2021-02-24
Rename and standardize test file names

jbethanco:
If we keep test file names consistent with the files that they test...
When in the file needing testing, we can press Ctr+Cmd+Up arrow and it will show associated interface file and if Ctr+Cmd+Up is pressed again the associated testing file.
So basically all we have to do for this functionality is name our test files the same name as our original files with the suffix "Tests" e.g. "YourMom.swift" has a testing file named "YourMomTests.swift"


bmorris-revacomm (approved):
Cleared for takeoff!

PR #256, opened 2021-02-24
Clean up some use of the takeoff date data.

Pete-Hoch:
Changed to display the h ours on the Sortie card and update the Date on the Events card.

Also removed some of the Date+Ext functions that are no longer used.


PBLogging/Views/Overview/EventCard.swift
```diff
@@ -16,7 +16,7 @@ struct EventCard: View {
             destination: RecentSortiesList(event: event),
             label: {
                 HStack {
-                    EventLabel(event: event)
+                    EventLabel(event: event, sortie: event.sorties.first ?? Sortie())
```
jbethanco, line 19: Adding a parameter for the sortie adds unnecessary complication. The first sortie is already passed in with the event.

PBLogging/Views/Overview/EventCard.swift
```diff
@@ -43,7 +41,7 @@ struct EventCard: View {
                         .resizable()
                         .scaledToFit()
                         .frame(height: 16)
-                    if let date = event.sorties.first?.takeoffTime {
+                    if let date = event.sorties.count > 0 ? sortie.takeoffTime : nil {
```
jbethanco, line 44: This has the same effect as the original code. 

Also, the date here was replaced in design with the event summary.

PBLogging/Extensions/Date+Ext.swift
```diff
@@ -6,64 +6,15 @@
...
-        let decMinutes = Double(minutes)/60.0
-        let total = Double(hours) + decMinutes
-        
-        return String(format: "%.1f", total)
-    }
-    
     func string24HourTime() -> String{
         let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "HHmm"
+        dateFormatter.dateFormat = "HH:mm"
```
jbethanco, line 11: Military Zulu time is typically written without the colon.

jbethanco (commented):
Please see comments.

jbethanco (changes requested):
Please see comments for changes.

PBLogging/Views/Overview/EventCard.swift
Pete-Hoch: it's not exactly the same, since first is an optional, we do not know that we will have a sortie or not. But the event cart needed a Sortie as an ObservableObject so that changes to it would redraw the EventCard.

But your next comment makes this immaterial as we will no longer display the date here.

PBLogging/Views/Overview/EventCard.swift
Pete-Hoch: As stated above, this is not the same as the original code did not push a redraw when the Sortie data changes. 

I had not seen the change in the Figma design. I'll update. Thanks!

PBLogging/Extensions/Date+Ext.swift
Pete-Hoch: Thanks, that was my civilian background creeping in again.

PBLogging/Views/Overview/EventCard.swift
jbethanco: Event was an ObservedObject so changes would redraw the event card already. Even on its sortie properties.
This was unnecessary. 

If it were required to be its own observable object, it should have been established in an initializer, rather than as a second parameter.

PBLogging/Views/Overview/EventCard.swift
jbethanco: Please run current 0.2.0. If you change the date. It redraws. Perhaps you were on an older commit before it was changed to an Observed object?

PBLogging/Views/Overview/EventCard.swift
jbethanco: Passing in Sortie() doesn't rid us of the optional first. It just passes a sortie in with a guaranteed nil first that we have to check anyway.

Pete-Hoch: Well, maybe it should have updated, but it was not. Does not really matter anymore as there is no longer a date on the card.

PBLogging/Views/Overview/EventCard.swift
Pete-Hoch: Maybe so, but the date is gone now. So it really doesn't matter anymore.

Pete-Hoch:
Requested changes have been made.

jbethanco (approved):
👍

PBLogging/Views/Overview/EventCard.swift
jbethanco: Sorry. I think I see where it may not have been updating now. 
That might be a larger problem we may have to deal with. 
On a side note, instead of passing an object and and objects property into a View, you can write an initializer such as 
```
    @ObservedObject var event: Event
    @ObservedObject var sortie: Sortie
    init(event: Event){
        _event = ObservedObject(wrappedValue: event)
        _sortie = ObservedObject(wrappedValue: event.sorties.first ?? Sortie())
    }
```
It is safer and keeps the call site cleaner as you only have to pass in the event.

Pete-Hoch: True, that is cleaner when (if) we need to do it in the future.

PR #258, opened 2021-02-25
Bring unit tests for String+Ext.swift up to 100%

Pete-Hoch:
I found a nice way to add testing of calls to fatalError and have added it as well as testing of every path in String+Ext.swift.

I also had to restructure charactersAfterLastOccurrenceOf a little because there was no way to hit both nil return lines.


jbethanco (approved):
Nice. Fatal error thing is cool.

PR #259, opened 2021-02-26
Test opening SubmitAProblem.

Pete-Hoch:
Small test, but it brings out unit test coverage to 79.6%!


PBLogging/Views/UtilityViews/BoldText.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 struct BoldText: View {
     let text: String
     let size: CGFloat
-    var color: Color = .pblForegroundPrimary
```
jbethanco, line 13: I don't understand the change here. If you want the primary color, one can just leave color off of the initializer.  This makes the call site longer.

jbethanco (approved):
Works good here.

PBLogging/Views/UtilityViews/BoldText.swift
Pete-Hoch: We did not have anyone using the default setting for color. (Other than the preview code.) So this was effectively unused code.

PR #261, opened 2021-02-26
Add 781 and PDF Printing Sharing

jbethanco:
Add Generate PDF button to the top of the list of sorties in an event. (Need design for where this should go)
Add multi day FEMA mission to the Sample Data
Add extension on Event to generate Form 781s and PDFs based on day or serial number changes.
Bring back the PDFPreviewView, but have it handle multiple 781s.
Lots to refactor, but Christian wanted this for a VIP mission soon.


Pete-Hoch (approved):
Looks good.

I don't see comments being filled in yet. And should selecting some Sortie Types change anything on the 781? I don't see that either.

What affects the sort order of the Aircrew? The 781 does not match the app. It probably doesn't matter, but maybe they should be the same in both places?

jbethanco:
> Looks good.
> 
> I don't see comments being filled in yet. And should selecting some Sortie Types change anything on the 781? I don't see that either.
> 
> What affects the sort order of the Aircrew? The 781 does not match the app. It probably doesn't matter, but maybe they should be the same in both places?

That's something I want to fix. It currently has the comparable logic for the FlightAuthDutyCode in both the AircrewData struct and on our CrewLine class. Would like to figure out a way to get them both in the same place. Not sure if that is causing it, or if you are looking at the main page on sortie... where I don't think it is sorted yet. 

Concerning sortie types... I don't think those effect the 781.

PR #262, opened 2021-02-26
PBS-682 Hookup Airdrop on Mission Sortie Info tab to core data

Pete-Hoch:
Do Int validate like SortieFuelEditFields does float for number input.


PBLogging/Utilities/EnforceValues.swift
```diff
@@ -0,0 +1,26 @@
...
+func enforceAndUpdateInt32(_ item: inout Int32, with stateObject: inout String){
+    stateObject = stateObject.enforceIntNumber()
+    if let integerValue = Int32(stateObject) {
+        item = integerValue
+    } else if stateObject.count == 0 {
+        item = 0
+    }
+}
+
+func enforceAndUpdateInt16(_ item: inout Int16, with stateObject: inout String){
```
jbethanco: These methods need a namespace/home. I think an extension on View would be appropriate so they can be called from any SwiftUI view.

jbethanco (approved):

jbethanco (approved):
Nice.

PBLogging/Utilities/EnforceValues.swift
Pete-Hoch: OK, I was not sure we wanted to limit them to only View use, but that makes sense as I doubt we will be validating the json data in the same way.

PR #263, opened 2021-03-01
Fix/training events scrollview

bmorris-revacomm:
- fixed scrollview so that it is spaced correctly and works properly


jbethanco (approved):
So much nicer. 
Nice job.

Pete-Hoch (approved):
Spectacular!

PBLogging/Views/TabViews/TrainingEventsTab/TrainingEventsView.swift
```diff
@@ -45,40 +58,82 @@ struct TrainingEventsView: View {
...
         }
         .onDisappear{
             dataController.save()
         }
     }
+    
+    func offsetChanged(point: CGPoint) {
+        yOffset = point.y
+        xOffset = point.x
+        print(point)
```
Pete-Hoch, line 133: We probably don't need this printing though.

PR #265, opened 2021-03-01
PBS-679 Add comparable function for flightAuthDutyCode

Pete-Hoch:
Add struct dutyCodeCompInfo and refactored comparable code from both CrewLine and AircrewData into one func.


PBLogging/Data/EntityExtensions/CrewLine+Comparable.swift
```diff
@@ -8,9 +8,39 @@
 extension CrewLine: Comparable {
 
     public static func < (lhs: CrewLine, rhs: CrewLine) -> Bool {
+        return dutyCodeLessThan(dutyCodeCompInfo(lhs), dutyCodeCompInfo(rhs))
+    }
+
+    public struct dutyCodeCompInfo {
```
jbethanco, line 14: Struct name should be capitalized.

jbethanco (approved):
Not sure why but the sorting isn't matching up on the forms and on the 

In AircrewListView changed line 65 to:
ForEach(sortie.crewLines.sorted()) { crewLine in

The sort still isn't matching the generated 781.

Not sure what is/isn't happening.

PBLogging/Data/EntityExtensions/CrewLine+Comparable.swift
Pete-Hoch: Opps, fixed.

Pete-Hoch:
> Not sure why but the sorting isn't matching up on the forms and on the AircrewListView.
> The sort still isn't matching the generated 781.
> Not sure what is/isn't happening.

Looking at SpaceX Recovery and the first Sortie, I see the sorted Aircrew as the same. So is it just some?

jbethanco:
> > Not sure why but the sorting isn't matching up on the forms and on the AircrewListView.
> > The sort still isn't matching the generated 781.
> > Not sure what is/isn't happening.
> 
> Looking at SpaceX Recovery and the first Sortie, I see the sorted Aircrew as the same. So is it just some?

Not sure, did on FEMA mission. We should probably write a test on CrewLine vs AircrewData, using the same data and then sort each and compare arrays just to be certain. I'm wondering if in my rush to get out the 781 generation I forgot to copy over names or something into the aircrewdata from the crewlines.

PR #267, opened 2021-03-03
Rename 781 Struct items to align with Sortie, Crewline, Person names

Pete-Hoch:
Covers PBS-781

Make 781 struct property names congruent with Sortie, CrewLine, Person property names as much as possible.


jbethanco (approved):
Looks good.

PR #268, opened 2021-03-03
Fix/update event column

bmorris-revacomm:
- updated UI for training events view based on discussion with Rachael


Pete-Hoch (approved):
Looks nice.

PR #269, opened 2021-03-03
Fix/replace magic numbers

bmorris-revacomm:
- CGFloat extension for cornerRadius constant
- replaced all previous cornerRadius values
- cleaned some white space
- removed BoldText file because it was unused


Pete-Hoch (approved):
Nice!

PR #270, opened 2021-03-03
replace backgrounds with borders, refactor tab bar and tab views

bmorris-revacomm:
- refactor tab bar and tab views
- replace backgrounds with borders


jbethanco (approved):
It works.

PR #272, opened 2021-03-05
Sortie Flight Time Editing

jbethanco:
Add new flight time view that switches between Pilots and Loadmasters
* Needs feedback from customers. This gets it in their hands so they can play and tell us what they need
* Does not have design on it yet. Need to get functionality worked out first as functional changes break design sometimes

Add Auditing protocol for now, so we can pass in either the Pilot auditor or the Loadmaster auditor into the same view. Would like to merge them if possible after logic is worked out. 

Add extension on CrewLine for "isPilot" and "isLoadmaster"

Correct FEMA mission sample data Flight Auth Duty Code from "FPB" to "IPB" for one crewmember on multiple sorties.

Add extension on double for rounding to decimal places of length

Add PilotTimeAuditor and LoadmasterTimeAuditor view models that provide state objects for SwiftUI views. 
* saves to Core Data
* splits time evenly
* audits some data points

No testing needed on these until functionality is locked in.


Pete-Hoch (approved):
Looks nice and splits up the time. Is it just suppose to do Primary? There are numbers in Secondary, but they do not get divided.

I like the error noticed.

jbethanco:
> Looks nice and splits up the time. Is it just suppose to do Primary? There are numbers in Secondary, but they do not get divided.
> 
> I like the error noticed.

Just primary so far. Still learning how it should work. Found out today that loadmasters don't use secondary at all. Going to remove that column on theirs.

PR #275, opened 2021-03-08
add unit tests for some accessors & extensions

Pete-Hoch:
This is for PBS-631.

Not at 80% (not even close) but taking smaller chunks to try and avoid conflicts, etc.


PBLogging/Data/EntityExtensions/Sortie+Accessors.swift
```diff
@@ -86,7 +86,11 @@ extension Sortie {
     
     public var crewLines: [CrewLine] {
         get {
-            let set = crewLines_ as? Set<CrewLine> ?? []
+            // I usually do not like as!, but unless the type of crewLines_ change
+            // this can not fail. So we would never get to the default if we use as?
+            // Also, if the type of crewLines_ changes, I think it's better for this
+            // to fail instead of silently 'work'.
+            let set = crewLines_ as! Set<CrewLine>
```
jbethanco, line 93: This is probably an unnecessary change. The nil coalescing optional casting has the same effect.  crewLines_ type will always be an optional NSSet. Core data forces this on us and it will not change, even in a future update. The reason we are using a nil coalescing optional casting is NOT because of the possible type a  crewLines_ might become, but because it IS optional and there is a non-zero chance it may be nil, and we would want an empty array at that point and not a crash. This is in line with the purpose of the assessors file.  I think it is unnecessary, but if we are worried about Apple changing core data generated types on us, we could write some tests that just confirm that NSSets are NSSets.

LoggingUITests/LoggingUITests.swift
```diff
@@ -145,4 +145,13 @@ class LoggingUITests: XCTestCase {
 
         XCTAssertEqual(originalAircrewCount + 3, currentAircrewCount, "Aircrew were not added")
     }
+
+    func testBackgroundAppAndReopen() throws {
```
jbethanco, line 149: Just curious. What is the purpose of this test? What would be the negative case that our code could cause?

jbethanco (approved):

PBLogging/Data/EntityExtensions/Sortie+Accessors.swift
Pete-Hoch: I bounced back and forth on one (thus the large comment) and when we had this:

crewLines_ as? Set<CrewLine> ?? []

The [] was never executed. And not finding a way to make it happen, it looked more like unused code than anything. If  Set<CrewLine> does not cast to match the definition of crewLines_ I think we have a big problem, and passing back [] would hide the bug. So that is why I went in this direction.

LoggingUITests/LoggingUITests.swift
Pete-Hoch: As the comment mentions below, I wanted to make sure one UI test called sceneDidEnterBackground(). There is not much to test, although perhaps it can be a more useful test if the data store is changed (but not saved) and when we relaunch we check that the data change was saved. Also, if we ever start to save a user pref on what screen they were on so we can resort it on open, this would be a nice way to test that too.

PBLogging/Data/EntityExtensions/Sortie+Accessors.swift
jbethanco: [] is currently not executed but definitely a possibility as crewLines_ is optional. 
If we need to delete an entire crew for some reason, making crewLine_ nil is an easy way. 
Set will never not cast from NSSet. That will always work. The point is that crewLine_ is optional. If you force cast, that includes a force unwrap.  We don't want to force unwrap something that has the possibility of being optional. It's not about the casting, it is about the optionality.

jbethanco: YAGNI ? Maybe wait till it has something to test?

LoggingUITests/LoggingUITests.swift
jbethanco: YAGNI ? Maybe wait till it has something to test?

PBLogging/Data/EntityExtensions/Sortie+Accessors.swift
Pete-Hoch: Well, in the unit test I sent crewLines_ = nil and [] was never called. In fact, the test passes just fine after setting crewLines_ to nil. So, I'm not sure how to get the old [] part of the line to execute.

jbethanco: That is peculiar. Let me do some testing.

jbethanco: This is really bizarre. Try doing a dump on crewLines_ after setting it to nil. 
As opposed to  "- nil" it shows "Optional(Relationship 'crewLines_' on managed object (0x600002395e50) <Sortie: 0x600002395e50> " 
It seems that the @NSManaged wrapper on crewLines_  has some unexpected side effects.  Further down in the dump it shows crewLines_ as what appears to be an empty set. 

If we can never get an NSManaged NSSet to be nil, then optional casting has no point. 

I tried setting crewLines_ to an NSSet of strings which also got an exception. It Seems the @NSManaged wrapper is doing a lot more than meets the eye. 

If we can't set it to nil, as? will never nil coalesce. 
If we can't set it to something other than  a set of CrewLines, then as! will never produce a runtime error. 

So it is looking like @NSManaged is handling our nil case and any runtime errors we need. 
Might as well force unwrap as it will always succeed and never be nil. 

Of note, seems that the @NSManaged wrapper only has this effect on relationships. Testing sortie.comments_ has the expected behavior even with the @NSManaged wrapper.

PR #276, opened 2021-03-08
update UI to match Rachael's designs

bmorris-revacomm:
Training Events View:
- add column colors
- add column dividers
- fix spacing
- fix text colors/weight


Pete-Hoch (approved):
Looks real nice.

PR #277, opened 2021-03-09
Add lock buttons to Pilot Flight Time for testing

jbethanco:
Change FormTabView so it could have a calculated header based and sortie hours.
Delete a couple of comments and print statements


cbrech (changes requested):
If all locks are locked PilotAuditor line 128 errors (divide by 0).

jbethanco:
> If all locks are locked PilotAuditor line 128 errors (divide by 0).

Nice catch.

cbrech (approved):
Code looks good, not sold on the concept yet.

PR #278, opened 2021-03-09
Added initial demo data on first app load

cbrech:
Make it easier to show off the app and let first time users explore without having to make up data.

Also added more info to the existing SampleData.


jbethanco (approved):
Mile-high commit club badge earned. Shaking the iPad wasn't cool enough?

cbrech:
haha shaking it only works in debug mode. If you install via TestFlight its not in debug mode

PR #279, opened 2021-03-09
Fix/frame label

jbethanco:
Add custom modifier that provides border frame with text label. 

Update Mission Data section to more closely match design with the new frame modifier.


Pete-Hoch (approved):
Looks nice.

PR #286, opened 2021-03-10
PBS-710 Refactor CrewLine comparable for increased readability

jbethanco:
Add documentation 
Rename variables


petermisik (approved):
Looks good.

PR #287, opened 2021-03-11
Fix/updated edit button

petermisik:
Made the edit button a little more uniformed.  Added a button to pull from PBS


PBLogging/Views/Overview/ListView.swift
```diff
@@ -0,0 +1,43 @@
+//
+//  ListView.swift
+//  Logging
+//
+//  Created by Pete Misik on 3/11/21.
+//
+
+import SwiftUI
+
+struct ListView: View {
```
jbethanco, line 10: ListView is a pretty generic name. I've got a bad habit of doing that too. Maybe EventListView?

jbethanco (approved):
Runs good here.

PR #288, opened 2021-03-11
PBS-678 Refactor "MergeCrew" out of Event+Form781

jbethanco:
Add comments
Add more data points to the Form781
Add test for Form781 generation
Add test for PDF generation


petermisik (approved):
Looks good to me.

PR #289, opened 2021-03-11
Fix/refactor enforce and update

jbethanco:
Merge two integer functions into one generic function
Add test for enforceAndUpdateFunctions


Pete-Hoch (approved):
Looks great, I love the unit test.

PR #290, opened 2021-03-11
add padding in event list

bmorris-revacomm:
- added padding for event list
- added background for total time on sortie cards


jbethanco (approved):
Looks good... I think it may conflict with Misik's approved https://github.com/airmencoders/Logging-iOS/pull/287
Probably going to require some manual changes to make them mesh.

PR #292, opened 2021-03-12
create shadow modifier and add to event and sortie cards

bmorris-revacomm:
- create pblShadow modifier
- add shadow modifier to event and sortie cards
- update sortie card colors/border
- update previews


jbethanco (approved):
Looks good. no abort trap 6.

PR #293, opened 2021-03-12
combine sortie info tab view into mission data tab view

bmorris-revacomm:
- combine sortie info tab into mission data tab
- delete unused file
- rename and move sortie folder into MissionDataTab folder


jbethanco (approved):
Works nice.

PR #294, opened 2021-03-13
small keyboard entery test and core data create fetch and delete managed objects.

Pete-Hoch:


LoggingTests/CoreDataTests.swift
```diff
@@ -0,0 +1,369 @@
...
+//  We are not testing relationships here.
+//
+//  Created by Pete Hoch on 3/11/21.
+//
+
+import XCTest
+import CoreData
+@testable import Logging
+
+class CoreDataTests: XCTestCase {
```
jbethanco, line 17: Somewhere near 90% of this file may be unnecessary unless I'm missing something. A lot of it seems to test Apple provided API/Class behavior or NSManagedObject default behavior. We should assume Apple's classes/APIs behave as advertised and only test our modifications of them. 

 Please see comments here: https://github.com/airmencoders/Logging-iOS/tree/fix/more-unit-tests%2Bcomments 

All the "get[NSManagedObject]" methods can be replaced with a generic if still necessary. 

The 11 methods... getEvents(), getSorties(), getFuel(), etc ... could be replaced with:

func getManagedObject<N: NSManagedObject>() -> [N]  {
    let fetch = N.fetchRequest()
    let results = try! context.fetch(fetch)
    return results as! [N]
}

It then can be called so long as the caller is explicit with type:
e.g. let events: [Event] = getManagedObject()

Maybe not perfect, but something similar?
Any time you catch yourself writing a lot of similar looking code repetitively ... it may be a good time to look for a generic solution. 

Some of these tests seem to be testing class object equality. If that is the intent, then it should be the same with the code.

LoggingUITests/LoggingUITests.swift
```diff
@@ -152,6 +152,70 @@ class LoggingUITests: XCTestCase {
...
+        firstEvent.tap()
+
+        self.app.buttons["addSortieButton"].tap()
+        self.app.buttons.matching(identifier: "sortieCard").firstMatch.tap()
+
+        // left side text fields
+        var textField = app.textFields["MISSION DESIGN SERIES"]
+        enterAndTestFreeFormatText(textField, "C017A")
+
+        // Let's just do a few (for speed), but leave these here
```
jbethanco, line 167: If the tests are worth it, then we should include them. These seem to be confirming that Apple TextFields perform as expected. If our code isn't involved so much, we shouldn't worry about them. If our enforcement of things are tested, we could probably do it via unit tests vs time intensive UI tests.

PBLogging/Views/TabViews/MissionDataTab/MissionDataAndFlightSeqView.swift
```diff
@@ -34,16 +34,22 @@ struct MissionDataAndFlightSeqView: View {
             Group {
                 TextField("MDS", text: $sortie.mds)
                     .withBorderLabel("MISSION DESIGN SERIES")
+                    .accessibility(identifier: "MISSION DESIGN SERIES")
                 TextField("Serial Number", text: $sortie.serialNumber)
-                    .withBorderLabel("Serial Number")
+                    .withBorderLabel("SERIAL NUMBER")
+                    .accessibility(identifier: "SERIAL NUMBER")
```
jbethanco, line 40: Accessibility identifiers should only be used where we can't find another way to identify the item we are looking for. All of the textFields here can be found without the accessibility identifiers. Using them when not necessary clutters the view code. If the labels are dynamic or there isn't some sort of label like on scrollview, that is when we should use Accessibility identifiers. To find button or textfield identifying strings, set a breakpoint in the test and use "po app" or more specifically "po app.textField" etc.

jbethanco (changes requested):
Please see specific comments in this review and branch with comments here: https://github.com/airmencoders/Logging-iOS/tree/fix/more-unit-tests%2Bcomments

LoggingTests/CoreDataTests.swift
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        event.name = "Quite an event"
+        dataController.save()
+        waitForExpectations(timeout: 5, handler: nil)
+        XCTAssertFalse(context.hasChanges)
+        let events = getEvents()
+        XCTAssertEqual(events.count, 1)
+        let firstEvent = events.first!
+        XCTAssertEqual(firstEvent.id, event.id)
```
jbethanco, line 57: I'm trying to understand the intent here. Are we testing that they are the same object? If so, why not:
`XCTAssert(firstEvent === event)`
or
`XCTAssertEqual(firstEvent, event)`

jbethanco (approved):
Cool.

LoggingTests/CoreDataTests.swift
Pete-Hoch: Good point, changed to: XCTAssertEqual(firstEvent, event)

(And removed XCTAssertEqual(firstEvent.name, event.name) also.)

PR #295, opened 2021-03-15
update mission data UI

bmorris-revacomm:
- update mission data textfield color
- update chip UI and labels
- fix chip spacing
- fix frames for sortie info
- replace sortie info dividers with ThickDivider
- update ThickDivider color


jbethanco (approved):
Nice updates.

PR #296, opened 2021-03-15
Update Training Events and Denormalize Event Types

jbethanco:
PBS-705 Denormalize event types
PBS-729 Separate Event Groups by section
PBS-727 Training Event values not displayed when closing and reopening

* Delete TrainingEventsView and replace with TrainingSwitchingView which allows the viewing of Training events via the contained TrainingStepperView.
* Add view model for the Training Stepper View which extracts the code out of the view from the old TrainingEventsView
* Add view modifier that executes action when view disappears or the app backgrounded.
* Update sample data to include some loadmaster events and add crewPosition to the MissionEventTypes.
* Add FlightAuthDutyCode struct to provide a single source of truth for Flight Auth Duty code information.
* Event groupings are in, but need to figure out how to round the top and bottoms of rects for each group.

Discussion: Started with PB-705 “Denormalize Event Types”
That broke the Training Events View.
Friday meeting with customer it was brought up that the Training Events View didn’t scroll via the eventID column, so needed to update also. Would rather have made smaller PRs, but interrelated components made it cumbersome.


bmorris-revacomm (approved):
Daaaaamn look at those comments... looks good to me!

PR #299, opened 2021-03-17
Auto closing the Submit a Problem window - PBS-635

Pete-Hoch:
This is the first attempt, so there is probably more to explore, but I'd still like to get feedback.

We are using the change in the URL after the feedback is given to trigger the close.
Also, passing the @Binding / Binding<Bool> to two locations, so maybe there is a better way.

But this works, so what does everyone think?


jbethanco (approved):
Works nice.

cbrech:
Looks great

PR #300, opened 2021-03-17
PBS-772 Get rid of bottom line in aircrew list

jbethanco:
Add isLastRow boolean to AircrewRow
Update loop to set boolean true if is the last crewLine.
Don't show the divider if the row is last


cbrech (approved):
looks good

PR #302, opened 2021-03-18
Fix/remove aircrew scrollview

bmorris-revacomm:
- remove scrollView and accessibility identifier
- update UI test for adding aircrew


jbethanco (approved):
Works perfect.

PR #303, opened 2021-03-19
Enhance Gonkulate (or is it Gonculate?)

jbethanco:
Begin TDD of gonkulate 
Add multiple tests and test helper function
Randomize test order and add thread sanitizer to build scheme
Remove Lock Buttons and isLocked
Add refresh to auditor protocol

TimeRow
*Remove isLocked var
*Add FlightAuthDutyCode to TimeRow
*Add heavily commented static spread method to TimeRow 👀
*Add description computed var to TimeRow
*Add zeroAll, and zeroTime function to TimeRow


cbrech (changes requested):
Looks like it is prioritizing Sec over Pri time. I ran it on the FEMA mission and it produced all Inst for me and all Sec for DeGraff.

cbrech:
.2

jbethanco:
> .2
> […](#)
> Sent from my iPhone
> On Mar 19, 2021, at 15:30, jbethanco ***@***.***> wrote: ﻿ Looks like it is prioritizing Sec over Pri time. I ran it on the FEMA mission and it produced all Inst for me and all Sec for DeGraff. Still working on a lot of the logic. It's pushing FPs auto to secondary right now. If pushed to primary... would need to deduct X amount of time for ghosting. What should the default amount be for a single sortie? 0.1? 0.2? other? — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or unsubscribe.

image

If default is 0.2, that would make this form show 3.7 for the FPCC as this would be the result of two sorties with 0.2. Want to confirm that is still good. Also, want to confirm this particular form should show 0.2 in the "Other" column for the FPCC.

cbrech:
The technical answer is an IP with one copilot only has to ghost time if the IP logs a takeoff or landing. 

In this example the IP should have 4.1 instructor. 0 in all other columns. The FP should have 3.9 primary and .2 secondary.

jbethanco:
> The technical answer is an IP with one copilot only has to ghost time if the IP logs a takeoff or landing.
> 
> In this example the IP should have 4.1 instructor. 0 in all other columns. The FP should have 3.9 primary and .2 secondary.

Copy, I think I'm getting it. I think I need more (verified) forms.  

So...
if FEMAmission { ... } 😀

cbrech:
We fill Eval, then Inst, then Pri, then sec. Only other if we can't give someone hours in the first 4.

cbrech:
Also not sure how I remove my change request 😬

jbethanco:
> Also not sure how I remove my change request 😬

I'll handle. And push some new changes in a bit.

jbethanco:
> We fill Eval, then Inst, then Pri, then sec. Only other if we can't give someone hours in the first 4. 

Would that apply... If 2 EPs were on the same 3 hour flight with 2 MPs. The EPS split eval time at 1.5 each. Then they split Instructor time at 1.5 each?

PR #304, opened 2021-03-20
First pass at building dev app for ns.

Pete-Hoch:
This puts the archive and the iPad in the folder `~/ns-builds`.
I am not sure if `~` can be used for the upload command to nowSecure, so I used the fixed path "/Users/runner/ns-builds/" there.
I am not doing anything fancy, like archiving in a new folder each time, so what we upload to ns will be overwritten. But that can be added if we think we would want it.

I (obviously) have not tried pushing this to a /release/* branch yet.


cbrech (approved):

cbrech:
merged cause I have no way of testing locally

PR #305, opened 2021-03-20
PBS-762 Finish Flight Time Calculations Gonkulate

jbethanco:
Distribute time evenly between pilots upon gonkulation press
Add multiple tests and test helper function
Randomize test order and add thread sanitizer to build scheme
Remove Lock Buttons from Flight Time view
Add refresh to auditor protocol

TimeRow
*Remove isLocked var
*Add FlightAuthDutyCode to TimeRow
*Add description computed var to TimeRow
*Add zeroAll, and zeroTime functions to TimeRow


cbrech:
Looks like a good start. Over time we will need to find ways to make it smarter (looking at it someone logged an event, etc.)

cbrech (approved):

PR #307, opened 2021-03-21
Add DataProtection Capability Entitlement.

jbethanco:
Need to update the provisioning profile in developer.apple.com to match prior to this, so this can pass tests/ work.


cbrech:
Is this a result of the NowSecure scan?

jbethanco:
> Is this a result of the NowSecure scan?
I think the initial NowSecure scan caught it. 
Assuming it will be a NowSecure issue as the sqlite is readable with the app terminated.

cbrech:
Go ahead and turn it on. Let's see if it break the pipeline (I doubt it)

jbethanco:
> Go ahead and turn it on. Let's see if it break the pipeline (I doubt it)

image

👆 Github Actions PBL profile is showing Data Protection Enabled... not importing?

jbethanco:
> > Go ahead and turn it on. Let's see if it break the pipeline (I doubt it)
> 
> image
> 
> 👆 Github Actions PBL profile is showing Data Protection Enabled... not importing?

image
🤔

jbethanco:
> > Go ahead and turn it on. Let's see if it break the pipeline (I doubt it)
> 
> image
> 
> 👆 Github Actions PBL profile is showing Data Protection Enabled... not importing?

I think the problem is around here:

PR #311, opened 2021-03-23
Now Secure deploy, in test mode

Pete-Hoch:
I assume we want to merge this into the general deploy, but for not, I have left it as a manual trigger.

This script set built and uploaded Logger to Naw Secure at about 10:15 HT on Tuesday, 23 March.


cbrech (approved):
lets leave in manual run. The stinking thing takes almost an hour to run on their end and I don't want to overload it.

PR #312, opened 2021-03-23
Feature/add search bar

bmorris-revacomm:
- added search bar to training events view
- highlights an event when the event name or id contains the search text
- fixed two training event name strings


jbethanco (commented):
Works good.

jbethanco (approved):

PR #313, opened 2021-03-25
PBS-706 Add More Realistic Sample Missions

jbethanco:
Add sample Air Force 1 mission
Move sample missions to individual files in their own folder
Prevent loadmaster tab from showing up in flight time if there are no loadmasters
Add Sortie+Calculation function to count number of persons on a sortie at
a specified crewPosition
Add test for the above function

Add two arguments to the Debug scheme:

-com.apple.CoreData.SQLDebug 0
-com.apple.CoreData.ConcurrencyDebug 1

The former to increases CoreData debug information if set to 1 or higher.
It does nothing at 0.
The later to assert we access core data only from the thread it was created on.

We were violating this which can cause unexpected crashes... When accessing events on the
background thread for the pdf creation Event accessors were causing core data hits.
Tested throwing it in a DispatchQueue.main.async.
When doing this, noticed that pdf creation speed was increased a lot,
and it did not lock the UI. Leaving it in the dispatch.async  for now.

Modified testPDFPreview() so that it tests that a pdf is previewed.


bmorris-revacomm (approved):
looks good 👍🤩✈️🦄

PR #314, opened 2021-03-26
Add perform block and save to first load data

jbethanco:
TODO: add perform blocks anywhere we are working with the context.
While not required, it ensures that the context is only worked on in the thread that created it.

remove deprecated synchronize call
refactor stringly-typed key


Pete-Hoch (approved):
Looks good, works well.

PR #315, opened 2021-03-26
PBS-764 Enhance testing phase one

Pete-Hoch:
1) Add expectation messages to XCTAssertions where needed.

This is rather vague, so I mimicked what I cold find and added expectation messages practically everywhere. 

2) Verify relevant XCTAssert* functions are used (e.g. XCTAssertTrue, XCTAssertFalse, XCTAssertNil, XCTAssertEquals, etc)

This should be done. 

3) Standardize order of test value vs expected value if needed 
(e.g. XCTAssertEqual(a, expectedA) vs XCTAssertEqual(expectedA, a)

The vast majority of tests were already of the form XCTAssertEqual(a, expectedA) so I made them all that way.

These last two items in the Jira issue will be handled in phase 2.
4) Look into adding XCTContext.runActivity for added context for error messages. 
5) Set continueAfterFailure to false where appropriate


jbethanco (approved):
Nice work.

PR #316, opened 2021-03-26
Disallows third party keyboard

jbethanco:
To silence NowSecure warning.


Pete-Hoch:
Are we sure that we want to do this?
We should check to see if anyone uses third party keyboards.
We can also change the score in Now Secure if we are just worried about the report.

jbethanco:
> Are we sure that we want to do this?
> We should check to see if anyone uses third party keyboards.
> We can also change the score in Now Secure if we are just worried about the report.

EFB iPads are locked down. Doubt anyone is using a custom keyboard. If they are, they can contact us, and we can allow per identifier.

Pete-Hoch:
> > Are we sure that we want to do this?
> > We should check to see if anyone uses third party keyboards.
> > We can also change the score in Now Secure if we are just worried about the report.
> 
> EFB iPads are locked down. Doubt anyone is using a custom keyboard. If they are, they can contact us, and we can allow per identifier. "To silence NowSecure warning." was poor word choice. This is about a vulnerability.

Sounds good. Looking...

Pete-Hoch (approved):
Looks good & works great!

PR #318, opened 2021-03-29
PBS-766 - Fix font problem on Form781 takeoff time

jbethanco:
First takeoff time font was different than the rest of the form.


petermisik (approved):
Font works.

PR #319, opened 2021-03-29
Make Pilot time splitting more polite

jbethanco:
First two clicks of Gonkulate allow EP/IP non-greedy distribution where EP/IP do not take secondary time from less experienced pilots.
Third and fourth taps allow EP/IP to be greedy.

Update tests to match
Add testNonGreedyAndGreedy

Conversation driving change here: 
https://chat.il2.dso.mil/tron-air-force/pl/jzhsc73zepnkfrdnocwre7ojtr


cbrech:
Lets chat after stand up to make sure I understand

cbrech (changes requested):
Can we add some indication to the user of what type of gonkulate is going to happen when they press the button?

I am wondering if we are trying to do too much with one button and should provide some options? Especially a mechanism for the user to specify certain times and gonkulate the rest.

jbethanco:
> Can we add some indication to the user of what type of gonkulate is going to happen when they press the button?
> 
> I am wondering if we are trying to do too much with one button and should provide some options? Especially a mechanism for the user to specify certain times and gonkulate the rest.

sure

PR #321, opened 2021-03-29
Add string subscript to return character at specified index

jbethanco:
Add test for subscript extension
Modify call sites in CrewLine where this eliminates code
Add test for CrewLine+accessors where this is used


petermisik (approved):
Looks solid.

PR #323, opened 2021-03-30
Fix/improve pilot time split

jbethanco:
Add greedy and non-greedy buttons for flight time split. 
Show directional arrows on the buttons to indicate initial direction of time distribution
 (up the first column, down the second or reverse). 
Add lock buttons next to every time field. 
Locked fields are subtracted from total and left alone during time distribution. 
Add row total errors with violator names. 
Add isGreedy parameter to gonkulate function
Add isLocked to DoubleItem
Add isLocked to time splitting logic
Add canBeGreedy to Auditor Protocol
Remove second button if canBeGreedy is false
Adjust spacing in UI for lock buttons
Add LockToggleStyle to UI
Add TimeField to UI


cbrech (approved):
Looks good. In a future PR can you add the ability to edit/lock the ghost time as well.

jbethanco:
> Looks good. In a future PR can you add the ability to edit/lock the ghost time as well.

Sure. Would just need to understand the logic.

PR #325, opened 2021-03-31
Updated name to Submit Feedback

cbrech:
So that users can send good stuff too not just problems


petermisik (commented):
I like the verbiage.

jbethanco (approved):
Nice.

PR #326, opened 2021-03-31
Feature/pdf training events - PBS(765)

bmorris-revacomm:
- create swiftUI views for training events pdf
- create function for converting UIImage of view into PDF
- combine 781 pdfs with training event pdfs


cbrech (approved):

cbrech:
Looks like a good MVP

jbethanco (approved):
Works and the pdfs look nice. Seems the pdfs combining on the main thread start locking the UI. We will need to go back in and put them on a background thread at some point. 

The print outs look good. If you haven't done so, check out the printer simulator...
Download additional tools for Xcode 12. 
Inside you will find the printer simulator in the hardware folder.

They can be found here:
https://developer.apple.com/download/more/
https://download.developer.apple.com/Developer_Tools/Additional_Tools_for_Xcode_12/Additional_Tools_for_Xcode_12.dmg

PR #327, opened 2021-04-01
Complete and create Tests for "createForm781"

Pete-Hoch:
Removed unused code.
Added backup calculation for full stops in 781 generation.
Added Remarks to 781 generation.
Added more unit tests to validate number son the 781.

We are down to only a few fields onthe 781 that we do not fill in. This seems to be because we are not tracking this information. 

Form 781 Front Page - Unfilled fields

Box #	Tag
39		"maint_review"

40 a-f	"pilot_review_a", "pilot_review_b", "pilot_review_c"
		"pilot_review_d", "pilot_review_e", "pilot_review_f"

41		"sarm_review"

42		"arms_input_date", "arms_input_initials"
		"arms_audit_date", "arms_audit_initials"

43		"printed_name", "rank", "branch", "dsn", "signature"

Form 781 Back Page - All fields assigned


LoggingTests/Event+Form781sTests.swift
```diff
@@ -38,14 +38,206 @@ class Event_Form781sTests: XCTestCase {
...
+        for annotation in page.annotations {
+           if annotation.fieldName == "grand_touch_go" {
+               XCTAssertEqual(annotation.widgetStringValue, "3")
+           }
+        }
+        for annotation in page.annotations {
+           if annotation.fieldName == "grand_full_stop" {
+               XCTAssertEqual(annotation.widgetStringValue, "3")
+           }
+        }
```
jbethanco, line 224: Could probably group these in the same annotations loop.  Might reduce the test time.

jbethanco (approved):
Looks good.

LoggingTests/Event+Form781sTests.swift
Pete-Hoch: Good catch. (Sloppy copy/paste range error.)

petermisik (approved):
I like it.

PR #328, opened 2021-04-01
update keyboard defaults - PBS(851)

bmorris-revacomm:
- add keyboard defaults for SSN, Flying Org, Flight Auth Num, and Serial Number
- add auto-capitalization for mission data fields


cbrech (approved):
Can you also turn on caps lock for the Mission Data fields?

bmorris-revacomm:
> Can you also turn on caps lock for the Mission Data fields?

Even for Harm Location?

cbrech:
Yes

PR #329, opened 2021-04-02
PBS-835 Info carryover from previous sortie

Pete-Hoch:
Leah didn't methion these, but do we also want to bring missionNumber & missionSymbol forward?

@cbrech ^^^^


cbrech:
Yes we do

PR #331, opened 2021-04-06
PBS-831 new crew edit view with delete

Pete-Hoch:
This has the new UI we talked about in this mornings meeting.

Delete is still broken, but I wanted to get this out and get feedback on the screen changes while continuing to look into the crash when deleting an aircrew member.


jbethanco (approved):
If the save fix is working for you, i think it is good to go.

Pete-Hoch:
I had to add one more change and nil out the person_ in the crewline before delting the crewline. 

This works, but I am not sure if we now have a dangeling reference to something in the database.

jbethanco:
No need to nil the person. It's automatic as the delete rule on Crewline is at to nullify.

Pete-Hoch:
> No need to nil the person. It's automatic as the delete rule on Crewline is set to nullify.

Well if I don't do it we crash. So I'm not sure what that means.

PR #332, opened 2021-04-06
Fix/training events bug - PBS(840) & PBS(882)

bmorris-revacomm:
- embed name header in scrollview so content frame doesn't extend off-screen
- disable scrolling for scrollview
- specify height for name text to center it because disabling scrolling messed this up
- removed unnecessary stack view


jbethanco (approved):
🤘Nice.

PR #334, opened 2021-04-07
Fix core data crash

jbethanco:
Deleting and then adding an aircrew member was causing a crash.
While not 100% sure of the cause, this seems to fix it.

Correct autocapitalization.


Pete-Hoch (approved):
The crash is gone.

PR #335, opened 2021-04-07
wrap core data changes in .perform() calls

Pete-Hoch:
I didn't do the test functions (not sure that is really necessary), but I think everything else is covered.


jbethanco (approved):
works.

PR #336, opened 2021-04-08
replace List with ScrollView to work around bug in List Item highliht.

Pete-Hoch:
Web research indicates thatthis is an OS bug if the List is inside of a VStack.

So replaced the List with a ScrollView (since we are not using and List features).


jbethanco (approved):
Looks nice.

PR #340, opened 2021-04-12
Remove cookies and destroy session on subsequent login attempts.

jbethanco:


petermisik (approved):
I like clearing the cookie cache on failure.

cbrech:
Still getting errors on the simulator when I set debug to the prod url. Put a log dump in MM

cbrech:
I think there are two different things, the repeated login error (which this helps) and the calls timing out

jbethanco:
> I think there are two different things, the repeated login error (which this helps) and the calls timing out

Increasing the timeouts.

PR #341, opened 2021-04-12
fix/update UI

bmorris-revacomm:
- change color of event name/summary textfields
- rename ThickDivider to PBLDivider, remove frame, replace dividers in training events pdf view


jbethanco (approved):
Nice. 🤖

PR #345, opened 2021-04-13
Update EventsOverview.swift

cbrech:
For the initial submission


jbethanco (commented):
Thinking we should wrap in an #if DEBUG so we don't have odd local branches while development continues?

cbrech:
I think that might throw flags, planning on submitting to apple then immediately doing a PR to bring it back

jbethanco (approved):

PR #350, opened 2021-04-14
Update ContentHeaderView.swift

cbrech:
Remove the Share and Print icons from the top menu since we aren't using them and they do nothing.


petermisik (approved):
I was going to ask about these buttons.  I like the removal.

PR #354, opened 2021-04-14
async generation of pdf files - PBS-899

Pete-Hoch:
I went simple with this. Since the code data is read only for this operation, I just do all the core data access up front and then generate the forms and images in the background.

UI is very responsive. 

Note: There is no reason to leave these as part of Events, but I didn't have a better place in mind to put the static methods. Open to suggestions if you think it's a big deal.


cbrech:
Not seeing the Training Events pages in the PDF preview anymore.

PBLogging/Data/EntityExtensions/Event+Form781s.swift
```diff
@@ -322,31 +322,22 @@ extension Event {
...
-                
+                completionHandler(.success(savePath))
             } catch {
                 print("Complete Failure.")
-                DispatchQueue.main.async {
-                    completionHandler(.failure(.couldNotFillOut))
-                    return
-                }
+                completionHandler(.failure(.couldNotFillOut))
             }
```
jbethanco, line 338: If this completion handler updates the UI it should be called on main.

PBLogging/Views/UtilityViews/PDFPreviewView.swift
```diff
@@ -64,10 +64,27 @@ struct PDFPreviewView: View {
...
+                                                                       trainingTotals: trainingTotals)
+
+        Event.generatePDFs(forms: forms) { result in
                 switch result {
                 case .success(let url781):
-                    PDF().pdfFromImages(filename: "TrainingEvents", images: event.createCrewMemberTrainingEventImages()) { result in
+                    PDF().pdfFromImages(filename: "TrainingEvents", images: trainingImages) { result in
                         switch result {
                         case .success(let trainingEventsURL):
                             combinePDFs(urls: [url781, trainingEventsURL])
```
jbethanco, line 90: I think we need a callback from combinePDFs here as well so we can put it on the background as well.

PBLogging/Views/UtilityViews/PDFPreviewView.swift
```diff
@@ -64,10 +64,27 @@ struct PDFPreviewView: View {
...
+        let crewMembers = event.getCrewMembers()
+        let trainingTotals = event.getTrainingEventTotals()
+        let trainingImages = Event.createCrewMemberTrainingEventImages(crewMembers: crewMembers,
+                                                                       trainingTotals: trainingTotals)
+
+        Event.generatePDFs(forms: forms) { result in
                 switch result {
                 case .success(let url781):
-                    PDF().pdfFromImages(filename: "TrainingEvents", images: event.createCrewMemberTrainingEventImages()) { result in
+                    PDF().pdfFromImages(filename: "TrainingEvents", images: trainingImages) { result in
```
jbethanco, line 87: pdfFromImages function doesn't spawn a background thread, but is using the thread it is called on.

jbethanco (commented):
Each of the time consuming functions...
1. Generate 781s
2. Generate Training Events
3. Merge pdfs
I'm thinking each of them should spawn their own background thread and then callback to main for reusability and clarity.

PR #356, opened 2021-04-15
Move connectivity monitoring duties out of the network client.

jbethanco:
for Single Responsibility Principle


petermisik (approved):
Nice work on the utility.  Great update that should keep up trucking.

PR #359, opened 2021-04-16
Fix/add sample data - (PBS-706)

bmorris-revacomm:
- create fake sample data with all fields filled out with unique values (for verification purposes)
- add 3 real mission data samples
- add all to device shake
- update grouped sortie count to 6 instead of 5


jbethanco (changes requested):
Some sorties have no aircrew. Some aircrews are doubled up.
Lets get together, I'll show what I'm seeing.

jbethanco (approved):
Awesome work. This is so helpful.

PR #360, opened 2021-04-16
fix/update textfield UI - (PBS-924)

bmorris-revacomm:
- refactor mission data borders
- add textfield for special use
- add steppers for touchAndGo & fullStop
- update static aircrew list font
- update font for event name/summary textfields and move them into scrollview
- add reserve status to static aircrew list


jbethanco (approved):
Looks good here. We probably need to update the header labels in static crew list in the future so they aren't truncated.

PR #362, opened 2021-04-19
Fix/design for download events

petermisik:
Ready for review.


PBLogging/Network/PBSEventCard.swift
```diff
@@ -0,0 +1,192 @@
...
+                                )
+            .padding(10)
+        }
+        .padding()
+        .frame(maxWidth: 600)
+        .frame(maxHeight: .infinity)
+        .background(Color.pblBackgroundDefault)
+    }
+    
+    func calculateTripTime(pbsEvent: PBSEvent) -> Double{
```
jbethanco, line 73: We have a private function "timeBetweenDatesFor781" in sortie + calculations. I'm thinking maybe we should bring that out to the public so we can use here and there? As the 781 time calculation is different from regular math. An 8.1 here might show up as 8.2 after import. Not a huge deal, but something to think about.

PBLogging/Network/PBSEventCard.swift
```diff
@@ -0,0 +1,192 @@
...
+        
+        }
+        .padding(.leading)
+        .frame(maxHeight: .infinity)
+        .background(Color.pblBackground)
+    }
+
+    func pbsPersonnelLeft(crew: [PBSCrew]) -> [String]{
+        var names: [String] = []
+        for x in 0..<crew.count {
```
jbethanco, line 99: The left and right column functions are identical except for a !.  Perhaps merge them to a single function with a bool parameter to indicate even or odd? Or be crazy and write an generic extension on array that lets us do something like, let leftColumnNames = array.oddElements

PR #363, opened 2021-04-19
fix static aircrew list

bmorris-revacomm:
- fix static aircrew list from truncating header text


jbethanco (approved):
🤘And the award for most unique branch name goes to... ⭐️

PR #367, opened 2021-04-21
Fix contrast issue - (PBS-942) 🎉

bmorris-revacomm:
- select high contrast for pblBackground color (this fixes the contrast issue, but I have no idea why - selecting high contrast on any of the colors fixes everything, apparently 🤷‍♀️)
- remove unnecessary foreground color


jbethanco (approved):
Fixes the issue. 💡

PR #371, opened 2021-04-22
Feature/add sortie info pdf - (PBS-927 & 989)

bmorris-revacomm:
- create SwiftUI views for mission sortie info pdfs
- create func for generating images from those views
- refactor TE image generation func to account for more events than what will fit on a single page (for summary pages only)
- add page numbers to TE summary pdfs
- move TE pdf view file into UtilityViews
- refactor pblBorder modifier to allow changing color


PBLogging/Views/UtilityViews/PDFPreviewView.swift
```diff
@@ -67,13 +67,30 @@ struct PDFPreviewView: View {
...
+                                    }
+                                } else {
+                                    combinePDFs(urls: [url781, missionSortiesURL])
+                                }
+                            case .failure(let error):
+                                print(error.description())
+                            }
                         }
+                    } else {
+                        combinePDFs(urls: [url781])
```
Pete-Hoch, line 93: This may conflict with the background work out for review.

PR #376, opened 2021-04-27
Background PDF generation

Pete-Hoch:


cbrech:
Wow this is sooo much faster. 

Small potatoes that could be changed later. The text that changes looks a little bit like a button. Would be great to move it closer to the spinner.

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-04-27 at 14 12 04

PR #377, opened 2021-04-28
Add defaults for MDS, harm, issuing unit, and unit charged

jbethanco:
Use defaults upon new sortie creation
Default values are set any time a sortie is saved and the fields are not empty
The most recent saved sortie will determine the defaults.


petermisik (approved):
Awesome work.

PR #378, opened 2021-04-28
Make 0.0 disappear when textfield entered.

jbethanco:
Flight time and flight condition text fields when entered 0.0 clears out.
If the text field is empty when they leave it it turns into "0.0"


petermisik (approved):
Nice fix.

PR #379, opened 2021-04-29
Two minor bug fixes

jbethanco:
Add identifiers to 2 buttons for UITests for Xcode12.5
to prevent hang

Ensure alert always presented from the main thread


petermisik (approved):
Looks great.  All tests pass cleanly.

PR #381, opened 2021-05-01
Remove generate Form 781 button when no info

jbethanco:
Removes the generate button if the bare minimum
information is not available to generate the 781s.
Currently, it just needs a takeoff time

Also added a viewContext refresh so the event
overview page updates after editing sorties


cbrech:
Can we leave the button but have a pop over saying what the min to generate is?

jbethanco:
> Can we leave the button but have a pop over saying what the min to generate is?

What would you think about a subdued label that says, "To generate forms, add at least one sortie with a takeoff time."

It would be in place of the button, when not enough information to generate. 
Saves two taps and the button doesn't generate an alert instead of a pdf.

cbrech:
that works!

cbrech (approved):

PR #382, opened 2021-05-03
Added range to prevent negative steps on touch & go / landings stepper.

petermisik:
Small fix to add range to steppers to prevent negative touch and gos and landings.


jbethanco (approved):
Works.
Some of the numbers like, 20, get truncated.

PR #384, opened 2021-05-04
create PDFGeneratorTrainingEventsTests

bmorris-revacomm:
- create unit test for crew member training event pdf generation


jbethanco (approved):
It works. 
We may need to start writing tests for our tests. 🧪

PR #385, opened 2021-05-04
Fix/delete sorties - (PBS-1029)

bmorris-revacomm:
- add functionality to delete sorties
- clean up code and white space


PBLogging/Views/Overview/EventsOverview.swift
```diff
@@ -33,64 +33,59 @@ struct EventsOverview: View {
...
     var editButton: some View {
-        TextAndIconButton(text: editMode ? "Done" : "Edit",
-                          size: 12.0,
-                          icon: "square.and.pencil") {
+        TextAndIconButton(text: isEditing ? "Done" : "Edit",
+                          size: 14.0,
+                          icon: isEditing ? "" : "square.and.pencil") {
             
-            self.editMode = !self.editMode
+            self.isEditing.toggle()
```
jbethanco, line 40: If we wrap with "withAnimation" it might be less jarring. Might be worth doing here and on sorties.

PBLogging/Views/Overview/EventsOverview.swift
```diff
@@ -19,7 +19,7 @@ struct EventsOverview: View {
     
     var body: some View {
         NavigationView {
-            EventsListView().environment(\.editMode, .constant(self.editMode ? EditMode.active : EditMode.inactive))
+            EventsListView().environment(\.editMode, .constant(self.isEditing ? EditMode.active : EditMode.inactive))
```
jbethanco, line 22: Should environment be on the following line? 
EditMode.active/inactive could be .active/.inactive and still read clean.

jbethanco (approved):
It works. Made some minor comments.

PBLogging/Views/Overview/EventsOverview.swift
bmorris-revacomm: I'm not sure... I didn't write that line, it was already there 🤔 Where did you want to put it? 
I can try removing EditMode though!

PBLogging/Views/Overview/EventsOverview.swift
bmorris-revacomm: I tried this, but it doesn't seem to be doing anything? I noticed it doesn't do anything when adding sorties as well

PR #387, opened 2021-05-05
allow decimal place for airland/airdrop weight - (PBS-1028)

bmorris-revacomm:
- change airdropWeight and airlandWeight to Double instead of Int32 to allow 1 decimal place


jbethanco (approved):
Works good here. ⚡️

PR #388, opened 2021-05-06
Fix/test suite update

petermisik:
General tests


LoggingTests/CoreDataTests.swift
```diff
@@ -81,6 +81,67 @@ class CoreDataTests: XCTestCase {
...
     }
+    
+    func testFetchEventWithID() throws {
+            // Create an event and store the id
+            // run the fetchEventWithID function
+            // verify that it returns an event.
+            
+            let event = Event(context: context)
+            let eventUUID = UUID()
+            event.id = eventUUID
```
jbethanco, line 92: No biggie and doesn't effect the test at all... the event automatically gets an ID on awake, in the AllEntityAwake.swift extension file so no need to generate or assign one here, I think.

LoggingTests/CoreDataTests.swift
```diff
@@ -81,6 +81,67 @@ class CoreDataTests: XCTestCase {
...
+            
+            try! context.save()
+            
+            let result = dataController.fetchEventWithID(String(describing: eventUUID))
+            
+            XCTAssertEqual(result, event, "Didn't pull the correct event")
+            
+        }
+        
+    func testFetchWithEventIDRetunNil() throws {
```
jbethanco, line 102: Return misspelled. I probably wouldn't have noticed this if I didn't have the spell check turned on. 
Edit -> Format -> Spelling and Grammar -> Check Spelling While Typing 
Didn't even know that was a thing until I saw it in this video a couple of weeks ago: https://youtu.be/8kiR_QsQNPc?t=128

jbethanco (approved):
Looks good. Couple of minor little comments.

LoggingTests/CoreDataTests.swift
petermisik: I removed the lines and re-ran the test and got 
"Unexpected or improperly formatted UUID parameter with type Swift.__StringStorage, expected NSUUID or well-formed NSString (NSInvalidArgumentException)".
I replaced the lines and the test passes.

PR #391, opened 2021-05-11
Remove unused code

jbethanco:
Remove DutyCodeCompInfo struct
Remove snapshot() extension from View


petermisik (approved):
Needed removals.  Leaner, meaner, faster.

PR #392, opened 2021-05-12
Fix/update checker fix

Adam-Roney581:
Updated checkForUpdate() to use swift's compare functionality to compare versions
Decomposed checkForUpdate() to compareVersions() and generateAlert() separately
Added Unit tests for UpdateChecker


cbrech (commented):
Runs well!

PBLogging/Network/UpdateChecker.swift
```diff
@@ -45,39 +45,49 @@ class UpdateChecker {
                 print(#function)
                 if let self = self {
                     
-                    if let versionInfo = versionInfo {
+                    if let versionInfo = versionInfo, let actualAppVersion: String = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
```
jbethanco, line 49: minor thing... using guard let instead of if let here would reduce nesting.

PR #393, opened 2021-05-12
Feature/add fake progress bar

xijiewang4:
Removed progress strings and its implementations 
added fake progress bar and its implementations
no unit test need?


jbethanco (approved):
Looks good. Nice job.

PR #395, opened 2021-05-14
Fix/feedback form colors - (PBS-1144)

bmorris-revacomm:
- update feedback form colors when switching between light/dark mode


petermisik (approved):
Love the color scheme and the find on the permissions.

cbrech (changes requested):
Is there a way we can prevent the form from asking for camera stuff? I'm wary of adding the camera permission, I think that will complicate the cyber review process.

bmorris-revacomm:
> Is there a way we can prevent the form from asking for camera stuff? I'm wary of adding the camera permission, I think that will complicate the cyber review process.

Unfortunately, I don't think so. There doesn't seem to be a way to remove the "take photo" option from the menu without also removing the photo library option. What we could do is open the feedback form in a separate web browser instead of pulling it up in the app with webkit, but if we do that, we don't have a way to auto-fill with what version and build number they are on (unless we have a way to have jira allow auto-filling form fields with url parameters)

bmorris-revacomm:
removed camera permission request - going to leave this PR as just updating form colors

cbrech (approved):

PR #396, opened 2021-05-14
fix/remove auto correct - (PBS-1132)

bmorris-revacomm:
- disable auto-correct for flight auth duty code
- 🦄🎉🥳🤩👻🙀👄🦄


petermisik (approved):
Awesome! 🤘 🔥

PR #397, opened 2021-05-17
Fix/generate pdf link highlight fix

Adam-Roney581:
Moved the generate pdf navigation link into it's own list item so that it only highlights itself


jbethanco (approved):
Fixes it.  👍

PR #398, opened 2021-05-17
Fix/chevron padding sortie card

Adam-Roney581:
Moved trailing padding from the sortie card to the navigation link


jbethanco (approved):
Nice

PR #400, opened 2021-05-19
PDF Generator Test Speed Up

Adam-Roney581:
Changed pdf generator unit tests to use cancellable sink to wait for pdf to be generated. Results in faster test


jbethanco (approved):
Looks good.

PR #401, opened 2021-05-19
Feature/header bar UI

xijiewang4:
took out infobutton and moved it into a separate file
added PBL divider for every overview page
merged views to add into native navigation bar


bmorris-revacomm (changes requested):
Nice! Can we move all buttons that are not the info button down so they more closely match the figma design?

xijiewang4:
> Nice! Can we move all buttons that are not the info button down so they more closely match the figma design?

yeah i think so, right now they are part of the native navigation bar items but i think i will just need to put them in a separate view instead or apply padding on them somehow

bmorris-revacomm:
> > Nice! Can we move all buttons that are not the info button down so they more closely match the figma design?
> 
> yeah i think so, right now they are part of the native navigation bar items but i think i will just need to put them in a separate view instead or apply padding on them somehow

Nevermind, don't worry about it! I think we are actually going to keep it this way and do some small updates later

bmorris-revacomm:
> > > Nice! Can we move all buttons that are not the info button down so they more closely match the figma design?
> > 
> > 
> > yeah i think so, right now they are part of the native navigation bar items but i think i will just need to put them in a separate view instead or apply padding on them somehow
> 
> Nevermind, don't worry about it! I think we are actually going to keep it this way and do some small updates later

If you can just remove the dividers, I think you're good!

bmorris-revacomm:
It looks like there's still one divider left on the recent sorties list view!

xijiewang4:
> It looks like there's still one divider left on the recent sorties list view!

ok yeah I think that is the reason i added all those dividers in the first place, the List view added a divider, im gonna see if i can remove it or hide it

bmorris-revacomm:
> > It looks like there's still one divider left on the recent sorties list view!
> 
> ok yeah I think that is the reason i added all those dividers in the first place, the List view added a divider, im gonna see if i can remove it or hide it

Oooooh interesting haha. Sounds good, thanks!

jbethanco (changes requested):
After tapping an event and then returning, the "Event" title is gone. 


Simulator Screen Recording - iPad Pro (9 7-inch) - 2021-05-20 at 16 45 32

bmorris-revacomm (approved):
looks good!

jbethanco (approved):

PR #404, opened 2021-05-21
PBS-1047 Change "Event" entity/object to "Mission" entity/object

jbethanco:
Required many changes.
Call sites, comments, tests, identifiers, and tons of other places. 
Left PBSEvents as is, can rethink verbiage on that when API becomes available. 
Feel free to squash and merge as approver to get it out of the way as it may have conflicts/headaches with concurrent PRs.


petermisik (approved):
Looks good. Ton of changes, but it serve us in the future.

PR #405, opened 2021-05-21
PBS-968 Add Keyboard Shortcuts

jbethanco:
Add custom TabbedTextField component
Present keyboard with accessory view with arrows

Screen Shot 2021-05-21 at 8 09 38 AM


jbethanco:
Not sure if this is fully capable yet.
We may need to add additional properties that can be set from SwiftUI if we want to use it in places such as Flight Time.
Would also need to fix the logic so that it can skip numbered tabbed fields if they are not visible.

PR #406, opened 2021-05-21
Fix/swift lint autocorrect

jbethanco:
"fix" doesn't seem to work yet.
"autocorrect" does.
Is this the same for everyone else? 

Also updated to some recommended project settings.
I'm thinking it would be good if we are all using Xcode 12.5 (12E262). Thoughts?


bmorris-revacomm:
sounds good to me!

bmorris-revacomm (approved):

PR #413, opened 2021-05-26
Migrate user defaults from DefaultsManager to Defaults

jbethanco:
Add new Defaults struct that handles user defaults in a safer, simpler, and testable manner.

Discussion: This initial push leaves the old code in place to use during a migration period.
After a couple of weeks, we can delete the old code. 
If someone doesn't migrate prior to old code removal, worst that will happen is they will 
need to get through the passcode screen again. Would rather that not happen though.

Require new additions to defaults to have default values.
Require new additions to defaults have unique key values.

Update SwiftLint.yml to allow TODOs, so that only
expiring todos get warning.
Update SwiftLint build phase to lint and fix as fix was not
providing warnings alone

Add expiring todo and documentation on when to delete the
old DefaultsManager code. Set it for 6/14/21 as that should be enough
time for everyone to run once for the migration to happen


petermisik (approved):
Looks good.

PR #416, opened 2021-06-01
Decompose/Increase Testing to Update Checker

jbethanco:
Decompose UpdateChecker
Enable dependency injection for UpdateChecker
Add more Unit Tests to UpdateChecker
Rename functions and variables for readability
Fix Defaults dependency injection problem
Remove monitor from update checker
Remove monitor from update checker test
Remove various print statements
Delete test that was dependent on external server


petermisik (approved):
Looks great and works. 🤘

PR #417, opened 2021-06-02
Refactor Connectivity Monitor

jbethanco:
Flip logic in scene delegate to wait 1 second
before checking for connectivity prior to performing
the update check.

Add "mockFlakey" to connectivity monitor
for SwiftUI previews.

Add cancel call to Connectivity Monitor deinit for completeness

Make NWPathMonitor optional so nil may be passed in
for testing, and canvas previews.

Wrap "flakey" and canvas previews using them in "if DEBUG"


bmorris-revacomm (approved):
Looks good from what i can tell  🧽

PR #418, opened 2021-06-02
PBS-1206 PDF Write Fail Handling

jbethanco:
Add alert if combined PDF fails to write

In hopes of finding what is causing share/print buttons to be greyed out.


bmorris-revacomm (approved):
Way too much code to review for one PR. Please do smaller PRs going forward.

PR #420, opened 2021-06-03
Changed DatePicker timezone to be UTC

Adam-Roney581:
Updated the DatePicker to show UTC timezone. This should update the correct time everywhere else. 

I don't think this alter night hours since when checking if it's night time, it uses the UTC time zone anyways. Please correct me if I'm wrong.


jbethanco (approved):
works

cbrech:
Merging so I can use in an hour on this trip 😁

PR #421, opened 2021-06-03
Sorting crew properly. Will now delete the right index too

renicklopez:


jbethanco (changes requested):
We sort using the crewline comparable extension. Not only last name.  
I think a "viewContext.refresh(sortie, mergeChanges: true)" in the onDisappear might work to refresh the sort.

petermisik (approved):
It works.  
Tested with new crew members with different flight auth duty codes.

jbethanco (approved):

PR #423, opened 2021-06-07
Aircrew invalid flight auth code fix

Adam-Roney581:
Bug fix for aircrew with invalid flight auth codes. 

Previously if you added an aircrew with an invalid flight auth code it didn't know how to sort it and it would change its order in the list randomly. Because of this, it broke the index when deleting and aircrew. 

https://jira.il2.dso.mil/browse/PBS-1242


jbethanco (approved):
Nice.

PR #424, opened 2021-06-07
Refactor Border Label Modifier

jbethanco:
Add custom shape to draw the border
Remove colorScheme modifier as it is no longer needed


xijiewang4 (approved):
Looks great!

PR #425, opened 2021-06-07
PBS-1204 Fix P1 Consent Login Broken

jbethanco:
Change login test url to actual API endpoint.
Simplify LoginWebView code.

Discussion: The new P1 consent screen broke our fragile login.
Our original login tested if we could get to the SSO account page.
The P1 consent screen is processed per domain. Our API was
a different domain than what we were testing login against, so we were getting the consent screen back when requesting JSON.

This version asks for JSON and when it doesn’t get it, knows it is not logged in.  While still hacky, should be safer from breaking on future changes. In the end, we should probably request a keycloak app login with a callback schema we could use with an ASWebAuthenticationView.


petermisik (approved):
Works well. 🔥 🎸 🔥

PR #426, opened 2021-06-08
PBS-1241 Fix Date to String Conversion

mcgarrys:
Changes:
- Rename string() to toString() to be more explicit
- Set DateFormatter timezone in toString() to be 'UCT'

Additions:
- Helper Functions For Testing
	- toStringWithTimeZoneBehindUCT() 
	- toStringWithTimeZoneAheadOfUCT() 
- Unit Tests
	- testToString()
	- testToStringAfterSettingTimeZoneBehindUCT()
	- testToStringAfterSettingTimeZoneAheadOfUCT()

Discussion: 
The timezone was not set for the dateFormatter in the toString() function.
This resulted in the timezone defaulting to the local timezone for the device. 
When the function was used, the date would sometimes appear to be the day prior or the next day, 
depending on if a local timezone was behind or ahead of 'UCT'.
This could be seen on the Sortie Overview Card, as well as in the final 781 PDF.

This version addresses the timezone not being set, and tests a few different things. 
The first test verifies that the converted date is incorrect if a timezone has a difference with UCT that 
overlaps into the previous or next day.
The second and third tests verify that setting the device's default timezone has no effect on the date 
when using the function with the fix.


petermisik (approved):
Look great and works.  Like the explicitness of toString.

PR #427, opened 2021-06-09
updated generate 781s button

xijiewang4:
updated the generate 781s button into an independent button
note: using Emptyview() on Navigational Link to avoid active selection when the button is not pressed


jbethanco (changes requested):
Almost perfect. Check out the dark mode design:
https://www.figma.com/file/L0iJ4e3M6lcwi1UvUawoLs/PBL---Swift-UI-Designs?node-id=271%3A334
Foreground color should be black.

jbethanco (approved):
Cool.

PR #428, opened 2021-06-09
Sortie list scroll to the bottom 

Adam-Roney581:
This is a way to scroll to the bottom of the sortie list after adding a sortie. I think there is a better way to do it with ScrollViewReader and ScrollView, but that is only available in iOS 14. So I found a workaround in the meantime. I followed this blog/guide that uses an UIViewRepresentable/tableview to scroll to the bottom of the list. 

https://lostmoa.com/blog/ScrollListToRowInSwiftUI/


mcgarrys (approved):
Looks good and works. Think that's as good as it gets without iOS 14, I saw similar solutions online.

jbethanco (approved):
It works.

PR #433, opened 2021-06-11
Tab View to Nav Links

jbethanco:
Remove Tab Bar
Move tabbed items to nav links
Restructure files hierarchy to match
Add temporary Nav Links, that will be replaced shortly
Update UI Tests to match

Discussion: Moves Tab bar items into individual nav links as discussed yesterday at design sync.
The current nav links provide no additional information on the sortie overview page as the original tab items.
Added two new Jira tasks for creation of prototype nav links which will replace the nav links until final guidance from design. 

image


petermisik (approved):
Works and looks great! 🙌 🔥 🤘  🙌

PR #434, opened 2021-06-11
Added last 4 SSN limit for the textfield

renicklopez:


jbethanco (commented):
Typically we have been using the onChange extension for enforcement things. I think the ICAO uses it?

renicklopez:
> Typically we have been using the onChange extension for enforcement things. I think the ICAO uses it?

changed to use onChange

jbethanco (approved):
Looks good.

PR #435, opened 2021-06-11
Refactor FuelAirlandAirdropView and fix textfield bug.

mcgarrys:
Add ✍️
- createStateStringFromNumber<T>()

Refactor 💥
- FuelAirlandAirdropView init().

Fix 🛠
- Bug where the airRefuel and takeOffCenterOfGravity textfields were hooked up in FuelAirlandAirdropView.

Discussion:
Refactored the init() in the FuelAirlandAirdropView init(), so that all values being assigned
a State<String> now use a helper function to determine if the value should be an empty String
or an interpolated value.

Found and fixed a bug where two textfields were hooked up in FuelAirlandAirdropView, 
with the help of @jbethanco. Basically if you typed a value in one of the textfields, 
the same value would appear in the other and vice-versa.


jbethanco (approved):
Awesome. Way less cluttered. 🧼 🔥

PR #436, opened 2021-06-14
Feature/pbs 1269 update aircrew section

xijiewang4:
updated aircrew section
-created a file for two generic button views(transparent or colored background)
-added button to edit airecew with nav link
-removed nav link applied on the airecew table 

updated aircrew edit section
-moved and changed add and edit buttons
-removed navbartitle

changed navbartitle inline display


PBLogging/Views/UtilityViews/ContentButton.swift
```diff
@@ -0,0 +1,127 @@
...
+    let title: String
+    let backgroundColor: String
+
+    @Environment(\.colorScheme) var colorScheme
+    
+    init(isSelected: Binding<Bool>,
+         isTransparent: Bool,
+         imageName: String,
+         title: String,
+         backgroundColor: String? = "primary") {
```
jbethanco, line 24: stringly typed. Why not pass color? Does it make the call site ugly?

PBLogging/Views/UtilityViews/ContentButton.swift
```diff
@@ -0,0 +1,127 @@
...
+                        .padding(.trailing, 12)
+                }
+                .background(Color(backgroundColor))
+                .cornerRadius(5)
+            }
+        }
+    }
+}
+
+private struct Examplebutton: View {
```
jbethanco, line 71: B should be uppercase. Could nest inside the preview instead of here.

jbethanco (approved):
Looks pretty good. Nice job. The row heights seem a bit taller than design.

PBLogging/Views/UtilityViews/ContentButton.swift
xijiewang4: oops it was not even intended to be called Examplebutton

PBLogging/Views/UtilityViews/ContentButton.swift
xijiewang4: good point, i think i used string since i saw it was refer in color extension file

PBLogging/Views/UtilityViews/ContentButton.swift
xijiewang4: ok so it is unused code

PR #438, opened 2021-06-15
PBS-1206 Add filename check/conversion.

mcgarrys:
Add:
- convertToValidFileName() String extension

Discussion:
After doing some testing and research, it seems that "/" is the only character that 
will throw an error if it is in the file name when trying to save to disk. 

This function replaces all instances of "/" with "_", with the exception of instances 
at the beginning or the end of the string, which are replaced with "".


petermisik (approved):
It works. 🔥 🤘 👽

PR #439, opened 2021-06-15
PBS-1222 Add flight time check

mcgarrys:
Add:
- Text() to MissionDataView that says "Both take off and landing times are required in order to split flight time."

Discussion:
The above text would replace the "Split Flight Time" NavigationLink, if any of the following conditions are met:
- Take off time is not set.
- Landing time is not set.
- Flight time is less than 0.1 hours.

Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-06-15 at 16 30 20


cbrech:
Can we make this message an Alert instead (when someone clicks on the button)? That was we always show the split time button but it won't work unless the times are valid.

mcgarrys:
Yeah sure, will change that now.

cbrech (approved):
Looks good

PR #440, opened 2021-06-17
Sort Missions by date, add date to mission card, update sortie sort, remove scroll manager

Adam-Roney581:
- Add [date label to mission card](https://www.figma.com/file/L0iJ4e3M6lcwi1UvUawoLs/PBL---Swift-UI-Designs?node-id=1101%3A863). Shows the most recent sortie takeoff time. If there's no sortie/valid takeoff time, it shows the current date
- Sort missions by my most recent sortie takeoff time. If no sortie/valid takeoff time, show mission at the top of the list
- Updated sortie sort to show the most recent sortie at the top. If there's no takeoff time (or a new sortie) it shows at the top
- Removed scroll manager stuff since we're adding everything to the top of the list and sorting by most recent 
- Updated UI tests


PBLogging/Views/Overview/MissionCard.swift
```diff
@@ -26,6 +27,48 @@ struct MissionCard: View {
...
+        func getTakeoffTime() -> String {
+            // Today's date
+            let dateFormatter = DateFormatter()
+            dateFormatter.timeZone = TimeZone(abbreviation: "UTC")
+            dateFormatter.dateFormat = "d MMM yyyy"
+            
+            // If there's no sorties, return current date
+            guard !mission.sorties.isEmpty else {
+                return dateFormatter.string(from: Date())
+            }
```
jbethanco, line 58: I think this could be deleted and the method will have the same result.

PBLogging/Views/Overview/MissionListView.swift
```diff
@@ -19,24 +19,61 @@ struct MissionListView: View {
...
+            // Loop through sorties until you find a valid takeoff time (in case the first/first ones don't have a takeoff time)
+            var firstTakeoffTime: Date?
+            for sortie in first.sorties {
+                if let takeoffTime = sortie.takeoffTime {
+                    firstTakeoffTime = takeoffTime
+                    break
+                }
+            }
+            
+            var secondTakeoffTime: Date?
```
jbethanco, line 41: My tiny screen wouldn't let me select more.  
This should probably all  be abstracted away in a Comparable extension on Mission as opposed to code in a view.
However this seems overly complicated for what we are trying to accomplish. We may be able to simplify this a bit.  
If we conform mission to comparable, then we can change the fetch request to sort for us. That return the delete function back to normal as well. If you tried this and it didn't work for some reason, please let me know.

jbethanco (changes requested):

PBLogging/Views/Overview/MissionCard.swift
```diff
@@ -26,6 +27,48 @@ struct MissionCard: View {
...
+                Spacer()
+            }
+            .padding(.top, 10)
+            .padding(.leading, 10)
+            .frame(width: 150)
+            .foregroundColor(Color.pblForegroundSecondary)
+        }
+        
+        func getTakeoffTime() -> String {
+            // Today's date
```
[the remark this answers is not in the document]
Adam-Roney581: This has been removed

PBLogging/Views/Overview/MissionCard.swift
```diff
@@ -26,6 +27,48 @@ struct MissionCard: View {
...
+                    Text(getTakeoffTime())
+                        .font(.pblBold(size: 18))
+                    Spacer()
+                }
+                Spacer()
+            }
+            .padding(.top, 10)
+            .padding(.leading, 10)
+            .frame(width: 150)
+            .foregroundColor(Color.pblForegroundSecondary)
```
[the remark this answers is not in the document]
Adam-Roney581: Fixed

PBLogging/Views/Overview/MissionCard.swift
Adam-Roney581: This has been removed

PBLogging/Views/Overview/MissionListView.swift
Adam-Roney581: Added a new data model attribute like we discussed and it is now sorting by that. Also the delete function is back to normal

PBLogging/Data/EntityExtensions/Mission+Accessors.swift
```diff
@@ -18,6 +18,21 @@ extension Mission {
...
+                missionDate_ = newMissionDate
+                return newMissionDate
+            }
+            
+            return missionDate
+        }
+        set {
+            missionDate_ = newValue
+        }
+    }
```
jbethanco, line 35: As opposed to storing the date, which could return incorrect information if a sortie date is edited, this can be a computed property. So no need to store it in missionDate_ .  Also not sure if this will return the temporal first date. Let's pair on Tuple on this if you have the time.

jbethanco:
Sent a DM on MM. I'm thinking the most reasonable way forward is to add a non optional "date created" attribute on Mission and Sortie.

PR #441, opened 2021-06-17
Added fix action for ensuring UTC compliance when generating 781s.

petermisik:
Issue: When time was entered as 0050, the Calendar.current.dateComponents would register this as EST.
Fix:  - Removed Calendar.current
        - Added Calendar(identifier: .gregorian)
        - Set calendar timezone to UTC
 
See PBS-1312


jbethanco (approved):
Looks good.

PR #442, opened 2021-06-17
Fix/pbs 1290 klbs sa

tsanders88:
- no uitests, not sure if we're adding now or holding off
- can make a iconbutton component if necessary if we want to reuse for existing night hours button
- not sure if we want border highlight for warning/error, but figured we didn't want it for info so left it out entirely for now
- right now the popover text is colored based on warning level. can change back to default if desired or do a foreground/background based on severity as well


PBLogging/Modifiers/FrameWithLabelModifier.swift
```diff
@@ -23,8 +23,15 @@ struct FrameWithLabel: ViewModifier {
         }
     }
     
+    func highlightColor() -> Color {
+        return detailDisclosure?.severity == DetailDisclosure.Severity.error ? .red :
+            detailDisclosure?.severity == DetailDisclosure.Severity.warning ? .yellow : .pblForegroundPrimary
+    }
```
jbethanco, line 29: Nested ternary operations are hard to read. Consider a switch? Better yet, maybe a computer property off of the Severity enum?

PBLogging/ViewModels/DetailDisclosure.swift
```diff
@@ -0,0 +1,17 @@
...
+import Foundation
+
+struct DetailDisclosure {
+    enum Severity: CaseIterable {
+        case info, warning, error
+    }
+
+    var text: String
+    var severity: Severity
+}
```
jbethanco, line 17: If not iterating, no need to conform to CaseIterable.

PBLogging/Modifiers/FrameWithLabelModifier.swift
```diff
@@ -44,6 +51,23 @@ struct FrameWithLabel: ViewModifier {
...
+                            Button {
+                                showDetails = true
+                            } label: {
+                                Image(systemName: detailDisclosure?.severity == DetailDisclosure.Severity.info ? "info.circle" : "exclamationmark.circle")
+                            }
+                            .popover(isPresented: $showDetails) {
+                                Text(detailDisclosure!.text)
+                                    .padding()
+                                    .multilineTextAlignment(.center)
+                            }
```
jbethanco, line 66: As opposed to:
`if detailDisclosure != nil {`
You can use: 
`if let detailDisclosure = detailDisclosure {`

After that you don't have to force unwrap it or treat it as an optional. 

Ternary operator for the image system name can be avoided if the severity enum has computed properties.

PBLogging/ViewModels/DetailDisclosure.swift
jbethanco: Enums in swift can have computed properties and are pretty powerful. We could do something like:

```
struct DetailDisclosure {
    enum Severity {
        
        case info
        case warning
        case error
        
        var imageName: String {
            switch self {
            case .info:
                return "info.circle"
            case .warning:
                return "exclamationmark.circle"
            case .error:
                return "exclamationmark.circle"
            }
        }
        
        var color: Color {
            switch self {
            case .info:
                return .pblForegroundPrimary
            case .warning:
                return .yellow
            case .error:
                return .red
            }
        }
    }
    
    var text: String
    var severity: Severity
}
```

Then:
`Image(systemName: detailDisclosure?.severity == DetailDisclosure.Severity.info ? "info.circle" : "exclamationmark.circle")`

can become:
`Image(systemName: detailDisclosure.severity.imageName)`

And this:
`.foregroundColor(highlightColor())
`
Removes the need for the highlightColor function and changes the call site to:
`.foregroundColor(detailDisclosure.severity.color)`

It moves a lot of the logic to the background and makes the call sites in the view more declarative and readable.

PBLogging/Modifiers/FrameWithLabelModifier.swift
```diff
@@ -44,6 +51,23 @@ struct FrameWithLabel: ViewModifier {
...
+                            .offset(x: -5)
```
jbethanco, line 67: Prefer to use padding over offset if possible. The offset changes the drawing location without actual moving the item being offset. Can have some confusing side effects. In this case none.

jbethanco (commented):
Overall looks pretty good. Bringing the disclosure button with the popover message out to its own component may be useful elsewhere. 

Might want to check using this in both light mode and dark mode. The warning text in the popover is difficult to read with yellow on white.

PBLogging/Modifiers/FrameWithLabelModifier.swift
```diff
@@ -44,6 +45,13 @@ struct FrameWithLabel: ViewModifier {
                             .background(sizeView)
                         Spacer()
                     }
+                    if let detailDisclosure = detailDisclosure {
+                        HStack {
+                            Spacer()
+                            DetailDisclosureButton(detailDisclosure: detailDisclosure)
```
jbethanco, line 51: May want to rethink the structure of the DetailDisclosureButton and DetailDisclosure. The call site seems repetitive. 
I think if you remove the DetailDisclosure properties inside the DetailDisclosureButton. Then the call site would easier to reason about. e.g. 
`DetailDisclosureButton(text: "Test", severity: .info)`
instead of 
`DetailDisclosureButton(detailDisclosure: detailDisclosure)`

PBLogging/Views/SingleSortie/FlightTime/AuditingFlightTimeView.swift
```diff
@@ -121,15 +114,7 @@ struct AuditingFlightTimeView: View {
...
-    var popoverText: some View {
-        // swiftlint:disable:next line_length
-        Text("If able, this button populates your night hours (between the end of evening civil twilight and beginning of morning civil twilight) with those calculated based on interpolated positions and times at a 6 minute interval along a great circle line between the takeoff and landing airfields. The calculations assume a direct flight and may be inaccurate for circuitous routes.")
-            .multilineTextAlignment(.leading)
-            .frame(width: 700)
-            .frame(maxHeight: .infinity)
-            .font(.headline)
-            .padding()
-    }
+    var popoverText = "If able, this button populates your night hours (between the end of evening civil twilight and beginning of morning civil twilight) with those calculated based on interpolated positions and times at a 6 minute interval along a great circle line between the takeoff and landing airfields. The calculations assume a direct flight and may be inaccurate for circuitous routes."
```
jbethanco, line 117: Should be a let, I think? 
Also if you include the comment:      // swiftlint:disable:next line_length 
on line 116, it will disable the swiftlint warning for line length.

jbethanco (approved):
Looks pretty good.

PR #443, opened 2021-06-18
Fix training event pdf dates (PBS-1311)

bmorris-revacomm:
- change removeTime to toString
- delete removeTime func
- update testGetPersonTrainingEvents
- add UTC timezone to all sample data


PBLogging/Data/EntityExtensions/Person+Accessors.swift
```diff
@@ -9,12 +9,12 @@ import Foundation
 
 extension Person {
     
-    public func dailyTrainingEvents(for mission: Mission) -> [Date: [String: Int]] {
+    public func dailyTrainingEvents(for mission: Mission) -> [String: [String: Int]] {
         
-        var dictionary: [Date: [String: Int]] = [:]
+        var dictionary: [String: [String: Int]] = [:]
```
jbethanco, line 15: Test expecting Date instead of String. Tried "toString()" but now returns the wrong info. Weird.

bmorris-revacomm: Fixed it 👍

jbethanco (approved):
✓ Works
🛠 Fixes
🦧 Does stuff

PR #444, opened 2021-06-21
Fix/pbs 1338 file name conformance

renicklopez:


PBLogging/Views/UtilityViews/PDFPreviewView.swift
```diff
@@ -123,7 +123,7 @@ struct PDFPreviewView: View {
                                     return
                                 }
                                 let documentDirectory = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
-                                let outputFileURL = documentDirectory.appendingPathComponent("\(mission.name.convertToValidFileName()).pdf")
+                                let outputFileURL = documentDirectory.appendingPathComponent("\(mission.name.convertToValidFileName().removeSpecialCharsFromString()).pdf")
```
jbethanco, line 127: The two chained functions have the same intent, making the filename valid. Could we merge them into a single function? I think the call site could read much cleaner as:
`mission.name.safeFileNameString()  (or a better name)`
vs
`mission.name.convertToValidFileName().removeSpecialCharsFromString()`

We should probably make some unit tests on this to, to make sure it is doing what we think it is doing.

renicklopez: done

jbethanco (approved):
Works good.

PR #450, opened 2021-06-29
Combat/Combat Support hours and sorties for pdf generation

Adam-Roney581:
Added combat and combat support hours and sorties for pdf generation. 

Also added combat and combat support to the sortie type list for sortie info. 

Note: This is not saving the combat/combat support hours and sorties in the data model, it is just calculating them at pdf generation.


jbethanco (approved):
Works great. Nice job. Glad you didn't save the hours to the data model. We should probably remove those fields from the data model in the next data model version.

PR #451, opened 2021-06-29
patch to fix the crew delete crash reported by Christian.

petermisik:
Christian reported the app crashes when deleting personnel.  Here's the patch.

Fix: 
Changed from a range iteration on the ForEach Loop to an array/id combo.
Updated the line to pull the index from the array and see if it's even or odd and apply the styling.


mcgarrys (approved):
Looks good 💥

PR #452, opened 2021-06-30
Feature/export data model

bmorris-revacomm:
Initial step in the mission upload process
- Make extension files for all Data Models (except MissionRecordTypes) to make them encodable
- add upload button that prints the mission's json
- add unit test for printing json


jbethanco (approved):
Works.

PR #453, opened 2021-06-30
Fix/upload button

bmorris-revacomm:
- refactor upload button to present action sheet/gray it out
- adjust line lengths to prevent warnings
- there were two FlightTimeView files in Compile Sources - deleted one which apparently deleted both? so I had to delete the file from the project and make a new one lol


jbethanco (approved):
Nice.

PR #454, opened 2021-07-01
Feature/pbs 1025 tails

petermisik:
Summary:
     Pull in Tail number to populate serial number, mds, harm location, and unit charged.  Serial number text field now contains a drop down feature and as the user types, the list shrinks.


cbrech:
Screen Shot 2021-07-01 at 11 44 18 AM

I like it overall, couple things. 
Can you sort these?



Screen Shot 2021-07-01 at 11 46 22 AM
Can we add some placeholder text like "Type to search" or something like that

cbrech:
Can you make the rows a bit taller so they are easier to click on with a finger?

jbethanco (changes requested):
image
Really think we need some kind of border/drop shadow or something around the edges here. It just cuts into the view without providing much context for what it is autocompleting for. 

When used in dark mode... nothing shows up... but the border looks better.
image

jbethanco:
Also... Seems clicking outside the dropdown should have dismiss the list. It persists and a user must scroll around.
Simulator Screen Recording - -  _iPad Pro (9 7-inch) - 2021-07-01 at 21 16 34

petermisik:
Closing PR.  Will generate new one when refactoring is complete.

PR #457, opened 2021-07-01
Fix/reserve status - (PBS-1388)

bmorris-revacomm:
- only show reserve status text on 781 pdf and aircrew list if reserveStatus is not 0
- clean styling for Mission extension file


jbethanco (approved):
Nice cleanup too.

PR #458, opened 2021-07-01
PBS-1394 Update textfield placeholders

mcgarrys:
**Discussion**
As requested in PBS-1394, this commit updates the placeholder text to be "0" in the FuelAirlandAirdropView. When a user inputs "0", it will be replaced with the greyed out placeholder (now set to "0"), instead of clearing the textfield.

**Update** 🔄
- FuelAirlandAirdropView


bmorris-revacomm (approved):
Looks good! 🤩

PR #463, opened 2021-07-06
Fix/pbs 1391 fix flying org

petermisik:
Simple patch to streamline the flying organization.


mcgarrys (commented):
Overall looks good. One thing that I noticed is you are only applying the extractFlyingOrg extension when importing events. Is this something we should cover for user-inputted flying orgs?

E.g. Right now, if the user goes into a sortie and edits a crew member's flying org to be "16th Airlift Squadron", the extension isn't applied and it remains as such.

mcgarrys (approved):
Nice 👌 🔥

PR #464, opened 2021-07-06
Fix/Refactor String Ext funcs

bmorris-revacomm:
- refactored filtering strings to improve performance


jbethanco (approved):
Noice. Twice as fast.  💨

PR #465, opened 2021-07-07
Refactor full stop logic and clean formatting.

jbethanco:
Remove some overcomplicated full stop logic. Each sortie will have one full stop. This shouldn't ever change.

To Test: Change some touch and goes on various sorties and see if the generated Form781 counts make sense. Add some sorties, make sure the Full stops go up by one.


mcgarrys (dismissed):
Looks good and works 🤫 🔥. I did some tests and the pdf generates as expected. Since full stops will always be 1, should we remove the full stop stepper in the MissionDataView?

jbethanco:
> Looks good and works 🤫 🔥. I did some tests and the pdf generates as expected. Since full stops will always be 1, should we remove the full stop stepper in the MissionDataView?

I'm thinking so. Waiting on confirmation.

mcgarrys (approved):
I was dismissed 🥲 🙅‍♂️  Second time's the charm?

PR #467, opened 2021-07-08
Initial commit of clean up.  Fixed call site and function.

petermisik:
Discussion
Reworked PR-463 to remove fragile computed property that went from String -> Int -> String.  Cleaned up call sites.

Note: Put a similar property on the model to clean up the import conversion.

Add ✍️

None

Update 🔄

DataController
CrewLine+Accessors
String+Ext
ImportEventsView
PBSModels
AircrewListView


PBLogging.xcworkspace/contents.xcworkspacedata
```diff
@@ -4,4 +4,7 @@
    <FileRef
       location = "group:PBLogging.xcodeproj">
    </FileRef>
+   <FileRef
+      location = "group:Pods/Pods.xcodeproj">
+   </FileRef>
 </Workspace>
```
bmorris-revacomm, line 10: Can you see if you can find a way to remove/delete this reference to Pods?

petermisik: Removed.  Sorry about that.

bmorris-revacomm: No worries! Thank you!!

jbethanco (changes requested):
Can no longer edit existing flying orgs.

PBLogging/Extensions/Binding+Value.swift
```diff
@@ -0,0 +1,20 @@
...
+    func max(_ limit: Int) -> Self {
+        if self.wrappedValue.count > limit {
+            DispatchQueue.main.async {
+                self.wrappedValue = String(self.wrappedValue.dropLast())
+            }
+        }
+        
+        return self
+    }
+}
```
jbethanco, line 20: What is the purpose of the async call?

PBLogging/Views/SingleSortie/AircrewListView.swift
```diff
@@ -98,8 +98,8 @@ struct AircrewRow: View {
                     crewLine.person.last4 = crewLine.person.last4.enforcedSSNLast4()
                 })
                     .keyboardType(.numberPad)
-                TextField("FLYING ORG", text: $crewLine.flyingOrganization.onChange {
-                    crewLine.flyingOrganization = crewLine.flyingOrganization.extractFlyingOrg
+                TextField("0016", text: $crewLine.flyingOrganization.max(4), onEditingChanged: {_ in
+                    crewLine.flyingOrganization = crewLine.flyingOrganization.paddedFourDigitsOnly()
```
jbethanco, line 102: paddedFourDigitsOnly is now allowing alphas and digits, yet this function still is named with digits.

jbethanco (changes requested):
The Flying Org field now allows alphas and numbers, but now also allows punctuation which shouldn't be allowed. We need to limit it to alphanumerics only. When typing a 5th character, it briefly shows up then disappears.  

We are doing a similar but different validation on last4 for the SSAN on the same view, I think we can use a near identical solution for the flying organization.  

The current solution in this PR adds a file for another extension on binding that limits the data to 4 characters, but then assigns it a value that is passed to a function that limits the value to 4 characters. The binding.max(4) and the paddedFourDigitsOnly() are both trimming the string to length.

jbethanco (changes requested):
The paddedFourCharactersOnly code isn't getting called when the user is typing. Only after they leave the field which isn't our desired result. 

The function seemed it was doing redundant work with prefix(4) and suffix(4) but suffix is being called on self instead of s, which I don't think is the intent.

PBLogging/Data/EntityExtensions/CrewLine+Accessors.swift
```diff
@@ -86,4 +86,8 @@ extension CrewLine {
     public var isLoadmaster: Bool {
         flightAuthDutyCode[1] == "L"
     }
+    
+    public var organizationNumbers: String {
+        flyingOrganization_?.paddedFourDigitsOnly() ?? ""
+    }
```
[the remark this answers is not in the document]
petermisik: Removed this call as it's not needed.

jbethanco (changes requested):
image

There seems to be bug in the function. I think it may be on line 185 where it returns self instead of s. I'm not sure if that will fix it though. Are you testing data entry in the simulator or device for expected behavior?

petermisik:
I was testing in the sim. Didn't see this behavior. Will have an update later tonight.

petermisik:
Latest commit fixed the problem.  Tested and there seems to be no unwanted behavior.

jbethanco (approved):
It works!

PR #469, opened 2021-07-08
review at your own risk 😬💀

bmorris-revacomm:
- added a gazillion marks
- cleaned white space and formatting
- removed unnecessary Groups on previews
- removed some unnecessary extensions (PBLWebView, FlightTimeView, FlightConditionsView)


jbethanco (approved):
Should be useful. 🤔

PR #471, opened 2021-07-12
PBS-1422 Fix FlightTimeView cell heights

mcgarrys:
**Discussion** 🗣
This PR brings in some changes suggested by Christian after the initial UI update, as well as some things discussed the design/dev sync last week. 

Important Notes ✍️
- The FlightTimeView cell heights are now shorter, so the user can see more at once on the page. 
- The lock button for disabling textfields has been removed from inside of the textfield, as it was originally. 
- The whole view now adapts dynamically to portrait and landscape mode, instead of having constant widths. 
- The Gonkulate and Night Hours buttons are now in the bottom right of their respective sections, mimicking a traditional calculator app design.
- The column titles are now center aligned with the textfields in each row.
- The INS (P/I/E) and SIM INS (P/I/E) columns are now not shown if user is in the Loadmaster tab.

**BEFORE** 🙅‍♂️

Flight Time And Conditions View Demo 

**AFTER** 🤩

Updated Flight Time And Conditions View Demo


bmorris-revacomm:
good use of emojis 👍

jbethanco (approved):
Looks great. Code looks clean. Your PR format is putting the rest of us to shame. 

Couple of tiny things... 
image

1. If the user puts 10.0 or more, a couple of the columns truncate. We may be able to bring the side margins out a 1/2 inch to give them some more room? 
2. The gonkulate button seems vertically centered, maybe throw a spacer under so when the error list grows in height, the button stays pegged to the top of its container instead of traveling down the screen.

jbethanco (approved):
Works great.

PR #472, opened 2021-07-12
Fix/reset flight time and conditions - (PBS-1469)

bmorris-revacomm:
- create funcs for resetting flight time and flight conditions
- call resets in onChange for takeOffTime and landTime if totalTime changes


jbethanco (approved):
Solves the problem. Works great here.

PR #475, opened 2021-07-14
Removed lines 94-95 disabling setting date/time for first sortie. 

petermisik:
Discussion
Customer reported "I think it looks weird having the sortie times default to something valid. How does a user know they still need to fill it out? Maybe default to 0 flight hours so it’s obviously wrong or something else like that"

Add ✍️

None

Update 🔄

DataController

Removed lines that set the date/time when a mission is downloaded from Puckboard.


jbethanco (approved):
Works.

PR #476, opened 2021-07-14
Assign mission before takeoff date in sample missions

jbethanco:
-ensures they get the correct sorting date applied to mission


cbrech (approved):
Works

PR #478, opened 2021-07-14
Fix/update data model 791

bmorris-revacomm:
- updated core data properties for 791
- added the rest of the needed properties for 791+Accessors


jbethanco (approved):
Nice.

PR #480, opened 2021-07-15
Replace TextView with iOS 14 Text Editor

jbethanco:
Delete TextView file
Add UITextView appearance background clear to Content View
Replace TextView in CrewPosition SwitcherView and SortieCommentsView


bmorris-revacomm (changes requested):
Looks like it changed the font color, so I think just update that and it's good!

Edit: I also just noticed that there is no way to dismiss the keyboard unless you navigate away from the page. Even going into edit aircrew view and back doesn't work

jbethanco:
> Looks like it changed the font color, so I think just update that and it's good!
> 
> Edit: I also just noticed that there is no way to dismiss the keyboard unless you navigate backwards from the page. Even going into aircrew view/training events/flight time and then back doesn't work. I think it would be very confusing as a user to navigate forward, do work, but then have a keyboard pop up when I go back - I wouldn't know what it was for

Not sure how to handle it when they navigate away and come back, but I am able to dismiss the keyboard with the button in the bottom right corner, both in sim and on device.

bmorris-revacomm (approved):
Looks good! Did you change something with that bottom button? It's behaving differently for me now lol

PR #481, opened 2021-07-15
PBS-1453 Fix popover oddities

mcgarrys:
[PBS-1453](https://jira.il2.dso.mil/browse/PBS-1453)

**Discussion** 🗣
This PR addresses a few issues regarding the popovers used throughout the app. The main glitch being addressed occurs when a user tries to open a popover when one is already open. The intended popover will not show, and the popover that is already open will remain open. 

Upon doing some [research](https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/), I found that Apple strongly recommends that you only have one open at once. If you do need open a second popover, you must make sure the first one has been dismissed first. 

**Important Notes** ✍️
- Popover text no longer flashes black/white (depending on light/dark mode) when dismissing the popover
- Logic has been added to make sure that only one popover is open at a time. If no popover is already open, the intended popover is shown instantaneously. If one is already open, the intended popover will be shown after a brief delay, giving the other popover time to dismiss. 
- A longer delay will occur if the app is run on a simulator, to account for lag.
- If no delay is used, the original issue persists. The previous popover has to be completely gone from the UI before opening the next one.

**Testing** 
- Navigate to the Fuel, Airland, and Airdrop section in Mission Data. Input a very large number for any field pertaining to weight (KLBS), and try and tap on the different error buttons both slowly and rapidly. 
- Navigate to FlightTime view and alternate tapping the Gonkulate/Night Hours info buttons, both slowly and rapidly. 
- In both views, make sure there is only one popover open at a time, and there are no glitches occurring with the text (flashing, displaying the wrong text, etc.).
- Make sure this works in simulator and on device.


PBLogging/Components/DetailDisclosureButton.swift
```diff
@@ -34,17 +34,50 @@ enum Severity {
 }
 
 struct DetailDisclosureButton: View {
-    
+    @State private var showDetails = false
+    static var isPopoverShown = false
```
jbethanco, line 38: Since this is intentionally being used as a shared var across all instances, maybe naming "isAnyPopOverShown" or "isAnyDisclosureShowingPopover" or something better will share clearer intent.

PBLogging/Components/DetailDisclosureButton.swift
```diff
@@ -34,17 +34,50 @@ enum Severity {
...
+            if showDetails {
+                DetailDisclosureButton.isPopoverShown = false
+                showDetails = false
+            } else {
+                if DetailDisclosureButton.isPopoverShown {
+                    DetailDisclosureButton.isPopoverShown = false
+                    DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
+                        DetailDisclosureButton.isPopoverShown = true
+                        showDetails = true
+                    }
```
jbethanco, line 61: I'm normally a fan of verbose names, but in this case the length and repeated use of DetailDisclosureButton.isPopoverShown kind of clutters the reading. 

This is debatable, but using a typealias here could make things read easier... 
e.g. 
`typealias DDB = DetailDisclosureButton `
or even
`typealias D = DetailDisclosureButton`

"D.isPopoverShown" reduces the noise a bit and I personally think it makes it easier to read, even though the D would require an initial but quick understanding. I think the typealias could be private as well to prevent usage outside the struct.

PBLogging/Components/DetailDisclosureButton.swift
```diff
@@ -34,17 +34,50 @@ enum Severity {
...
+            /// This ensures any popover is turned off when isPopoverShown has changed, and is now false
+            if !DetailDisclosureButton.isPopoverShown {
+                showDetails = false
+            }
+        }
+        .onChange(of: showDetails) { _ in
+            /// This ensures that isPopoverShown is set to false, should a popover be closed by tapping anywhere on the screen besides it's respective button
+            if !showDetails {
+                DetailDisclosureButton.isPopoverShown = false
+            }
```
jbethanco, line 80: I think this could be...
`DetailDisclosureButton.isPopoverShown = showDetails`

and the same but inverse in the other .onChange

PBLogging/Components/DetailDisclosureButton.swift
```diff
@@ -34,17 +34,50 @@ enum Severity {
 }
 
 struct DetailDisclosureButton: View {
-    
+    @State private var showDetails = false
```
jbethanco, line 37: I don't think this was your var name choice, it kind of reads like a function name. 
https://github.com/airbnb/swift#bool-names

jbethanco (approved):
Works perfect. 🎯 Tested pretty heavy and I couldn't break it. Made some comments about some readability things, nothing major.  I doubt it would need it, but if you decide to make those or any changes, get mine or someone else's review before merge.

jbethanco (commented):
Nice

PR #484, opened 2021-07-16
PBS-1365 Incorporate Labels

mcgarrys:
[PBS-1365](https://jira.il2.dso.mil/browse/PBS-1365)

**Discussion** 🗣
This PR replaces HStacks containing Text() and an Image() with the Label view introduced in iOS 14. Labels can be very flexible (with things like element ordering, modifiers, custom views, etc.) when using custom Label Styles. 

**Important Notes** ✍️
- A custom Label Style called PBLLabelStyle() is used on almost all Labels added to the app, giving the ability to adjust font size, element order, and foreground color.
- Label() has default spacing between between it's Title and Icon, which is why I used an HStack in the custom Label Style.
- There were a few cases where I felt a Label didn't make sense, like within the ContentButton view (where the views are being created as vars). 
- If there are any places in the app where you feel we could use a Label() that I missed, let me know 🙂

**Resources** 📚
- [How to create and use Labels in SwiftUI](https://www.simpleswiftguide.com/swiftui-label-tutorial-how-to-create-and-use-label-in-swiftui/)
- [Label Styles](https://swiftontap.com/LabelStyle)


jbethanco (approved):
Looks tight. We need to make your PR format the standard. Nice job.

PR #486, opened 2021-07-19
PBS-1399 Incorporate onChange

mcgarrys:
[PBS-1399](https://jira.il2.dso.mil/browse/PBS-1399)

**Discussion** 🗣
This PR incorporates the native onChange modifier in places where we are observing values throughout the app (and performing logic when a change occurs). It replaces a custom extension on Binding that was being used prior. 

**Important Notes** ✍️
- Since onChange can be applied to any part or sub-part of the view that can access the observed value, I experimented moving all the onChange calls to the highest level HStack or VStack within the view. Though after trying, It seemed better for readability to keep the call site local (e.g. calling on Textfields, DatePickers, etc.).


bmorris-revacomm (approved):
looks good and runs! 🏃‍♂️💨

PR #487, opened 2021-07-19
Form 791 Font Size Change

jbethanco:
Change all fields font size to 12 on 791. 
Auto font size was causing text to not show up.


bmorris-revacomm (approved):
yayyy 🥳🤩

PR #488, opened 2021-07-19
PBS-1401 Incorporate LazyVGrid

mcgarrys:
[PBS-1401](https://jira.il2.dso.mil/browse/PBS-1401)

**Discussion** 🗣
This PR incorporates the native LazyVGrid introduced in iOS 14. LazyVGrid/LazyHGrid are powerful ways to display many of one type of element in a grid format, similar to Lists. For now, the only place it seems to be useful is with the Sortie Type choice chips in the Mission Data View.

**Important Notes** ✍️
- Sortie Type choice chips no longer have a fixed width, they adapt dynamically depending on device orientation. 
- Based on landscape designs in [figma](https://www.figma.com/file/L0iJ4e3M6lcwi1UvUawoLs/PBL---Swift-UI-Designs?node-id=3807%3A5466), the number of choice chip columns now changes from four to five when device is in landscape.
- When the view appears, instead of setting a number of columns based on the device's current orientation, it is set based on  the ratio of the view height vs the view width. This is to account for the case when a device's orientation is .flat, and the content can appear to be in either landscape or portrait.
- DeviceRotationViewModifer is used to detect a change in the device's current orientation. 

**Resources** 📚
- [Mastering grids in SwiftUI](https://swiftwithmajid.com/2020/07/08/mastering-grids-in-swiftui/)
- [Dynamic columns with LazyVGrids](https://stackoverflow.com/a/65422017)
- [Detect device rotation in SwiftUI](https://www.hackingwithswift.com/quick-start/swiftui/how-to-detect-device-rotation)


jbethanco (changes requested):
Found we could use Adaptive instead of Flexible with a similar output, but without the orientation code. 

https://github.com/airmencoders/Logging-iOS/tree/chore/PBS-1401-incoporateLazyGridAdaptive

I didn't bulletproof it or test heavy, but the branch above shows how it might be done.

jbethanco (approved):
Works good.

PR #490, opened 2021-07-20
Fix Form791 Attempt 2

jbethanco:
Discussion: Font size fix did not work.
Copied working fields and renamed them instead of making new fields.
This worked for one row in last form, hopefully works for all rows now.


bmorris-revacomm (approved):
🤞

PR #494, opened 2021-07-20
Feature/791 pdf form - (PBS-1389)

bmorris-revacomm:
- create file for generating 791 pdf
- create sample data for 791 mission (all 791 fields filled out)
- update pdf preview file to account for 791
- add more 791 fields to defaults
- call 791 defaults when sortie created (missed this before)
- removed unused property on 781 pdf generation file

Special thanks to John, without whom this would not have been possible 🏆


jbethanco (changes requested):
Works great here. 
Nice work! 🔨
But... needs the defaults write to the 791 , only if the 791 is used. So there is no 791 if they don't issue any gas.

jbethanco (approved):
Works nice... I think we still need to read from the defaults and write to the 791, if it is a new mission/sortie, but we can add later.

PR #495, opened 2021-07-20
Fix/json encoding - (PBS-1476)

bmorris-revacomm:
- create extension on Double to round and then covert to Decimal
- update all Double encodable calls with the extension


jbethanco (approved):
Fantastic.

PR #496, opened 2021-07-20
PBS-1495 Delete empty Issue/Aircrew rows

mcgarrys:
[PBS-1495](https://jira.il2.dso.mil/browse/PBS-1495)

**Discussion** 🗣
This PR adds functionality to remove empty rows (rows with no fields filled in) from the IssuesListView and the AircrewListView. The functions are called when navigating away from the views containing the lists. 

**Important Notes** ✍️
- The functions iterate backwards through the lists, as removing elements in a forwards order results in an index out of range error.
- One thing I noticed when testing: If you add an empty row to the Aircrew list and then navigate back to the previous view, you can see the empty row appear, and then quickly disappear from the static Aircrew list. Would be interested to hear thoughts on if we should address that or not. 

**Testing** ✅
- Add empty rows, ensure they disappear when navigating away.
- Add empty rows, and fill in only one field. Ensure the rows persist after navigating away and back. Repeat for all fields in the row. 
- Do this with both the Issues list and the Aircrew list.


PBLogging/Views/SingleSortie/AircrewListView.swift
```diff
@@ -82,6 +85,18 @@ struct AircrewListView: View {
...
+        for i in stride(from: sortie.crewLines.count - 1, through: 0, by: -1) {
+            if sortie.crewLines[i].person.lastName.isEmpty &&
+                sortie.crewLines[i].person.last4.isEmpty &&
+                sortie.crewLines[i].flyingOrganization.isEmpty &&
+                sortie.crewLines[i].flightAuthDutyCode.isEmpty &&
+                sortie.crewLines[i].reserveStatus == 0 {
+                deleteSelectedAircrew(offsets: IndexSet(integer: i))
+            }
+        }
+    }
```
jbethanco, line 99: This may be a lot of logic in the view (same with the 791). I'm wondering if adding an "isEmpty" accessor on crewLine and 791 may be beneficial. Then maybe functions on sortie to deleteEmptyCrewLines() and deleteEmptyIssues(). 

As opposed to the stride, we might be able to do something like:
`form791.issues.removeAll(where: {$0.isEmpty} )`

mcgarrys: Sounds like a great idea 🔥

jbethanco (approved):
Works great.

PR #497, opened 2021-07-22
Fix/list UI

bmorris-revacomm:
- update list padding
- update textfield placeholders
- add aircraft text


jbethanco (approved):
Looks nice.

PR #498, opened 2021-07-22
PBS-1216 Fix AC Commander sorting

mcgarrys:
[PBS-1216](https://jira.il2.dso.mil/browse/PBS-1216)

**Discussion** 🗣
This PR adds Aircraft Commander sorting functionality to all aircrew lists throughout the app, including the generated 781 PDF. If an ACC is set, that crew member will appear first in aircrew lists. Users will have the ability to set an ACC when editing Aircrew. 

**Testing** ✅
- Change ACC in "Edit Aircrew", and ensure that the selected pilot appears first in all aircrew lists after navigating away.
- Ensure that the ACC appears first in the generated 781 PDF.

**Important Notes** ✍️
- Aircrew sorting based on Auth Duty Code still occurs, but if an ACC is set, they will always appear first in the list regardless of their position.
- When editing aircrew, only pilots can be selected as the ACC.
- Data model has been updated to add an isAircraftCommander attribute to the Crewline entity, and the model is now on version 4.

**UI Changes** 👁
- If an ACC is selected, a bold "ACC" acronym will appear next to their name.

IMG_0048

- There is now a dedicated "ACC" column in the "Edit Aircrew" section, where a star will appear next to pilots. Selecting a crew member's star will make them the current ACC.


jbethanco (approved):
Works great here.

PR #499, opened 2021-07-23
Fix/ScrollViewReader for TextEditors

bmorris-revacomm:
- add scrollViewReader to mission data view and flight time view
- add/call func to scroll to textEditor upon tap gesture
- update sortie delay button text


jbethanco (approved):
Works great.

PR #500, opened 2021-07-23
Add reset flight time/conditions tests

mcgarrys:
**Discussion** 🗣
This PR adds unit tests for the reset flight time and reset flight conditions functions added earlier this week. 

**Important Notes** ✍️
- In both tests, the functions are tested for when they receive a crew position as a parameter (flight time or flight conditions is then only reset for that specified crew position in the sortie), and for when one isn't passed in (flight time or flight conditions is reset for all positions in the sortie).


jbethanco (approved):
The tests, test what they say they test well. (And if you are a fan of arbitrary number celebration... PR #500 🎉)

PR #501, opened 2021-07-26
Update CrewLine+Accessors tests

mcgarrys:
**Discussion** 🗣
This PR adds/updates unit tests for the CrewLine accessors to increase code coverage and cover all getter/setters. 

**Important Notes** ✍️
- Tests should cover all custom getters and setters in CrewLine accessors, including those with guard statements that return a fresh instance of a struct, if one hasn't been assigned to the var yet (because the original property is optional).


jbethanco (commented):
Works well.

jbethanco (approved):
Works well.

PR #502, opened 2021-07-26
Fix/Add binding on change funcs - (PBS-1540)

bmorris-revacomm:
- create binding funcs for enforcing ICAO, SSN, and flying org on change
- refactor onChange call sites
- remove unused code
- rename funcs
- reorder String funcs


PBLogging/Extensions/Binding+Ext.swift
```diff
@@ -0,0 +1,41 @@
...
+    func enforceFlyingOrgOnChange() -> Binding <Value> where Value == String {
+        Binding(
+            get: { self.wrappedValue },
+            set: { newValue in
+                let correctedValue = newValue.enforceFlyingOrgFormatting()
+                self.wrappedValue = correctedValue
+            }
+        )
+    }
+}
```
mcgarrys, line 41: Had an idea for this, since those three functions are almost identical. Could create an Enum called StringFormat (or something along those lines), with cases for .ICAO, .SSN, and .FlyingOrg. Then could just pass in the formatting case to the function, and run through with a switch statement in one function. Something like 

Screen Shot 2021-07-26 at 3 27 06 PM

Not sure if it is a better approach though or if I'm missing something, just an idea.

mcgarrys (approved):
Looks great and works 🔥 💫  Left a suggestion/idea for Binding extension funcs.

PBLogging/Extensions/Binding+Ext.swift
bmorris-revacomm: I like it! @jbethanco Thoughts on this?

jbethanco: I tried to find a reason to not like it, but I can't. 
It's easy to reason about at the call site, it's extensible, it's nice.  
What do you guys think of a function / enum name change so that it reads...
func enforceChange(with: icaoFormatting)
Just a thought. 
For the enum, maybe lower case on the ssn and icao, to match most swift/apple type usage...
image

bmorris-revacomm: what about something more like onChange(enforce: icaoFormatting)? That way they know when it's happening, not just what it's doing

jbethanco (approved):

PR #506, opened 2021-07-28
PBS-1548 Fix swidtag build phase

mcgarrys:
[PBS-1548](https://jira.il2.dso.mil/browse/PBS-1548)

**Discussion** 🗣
This PR fixes the issue with the swidtag file not showing the current release version. The build script has been changed so that it updates the swidtag file with the marketing version (from Xcode), rather than using the latest tagged git version. The problem was that versions had to be manually tagged in git in order for the previous script to work. Now it should automatically update anytime the version number is changed in the project settings.


cbrech (commented):
Nice, can we all add the build number to it so something like v1.1.0-886

Thanks

mcgarrys:
> Nice, can we all add the build number to it so something like v1.1.0-886
> 
> Thanks

@cbrech Yeah we can add the build number. Is the "v1.1.0-886" format you mentioned what you want to be reflected in the swidtag file?

cbrech:
Yup

jbethanco (approved):
Working for the two of us.

PR #507, opened 2021-07-29
Feature/pbs 1025 import aircraft info

petermisik:
Discussion
Add AircraftSearchButtonModifier
Add FileManager+Ext
Add case to APIRequestBuilder 
Add getAircraftInfo to ImportEventsView
Add if statement to decode to handle file not found
Change variable name from ControlTower to AircraftController
Add updateSortieWith func to MissionDataAndFlightSeqView 

Add ✍️
AircraftSearchButtonModifier
FileManager+Ext

Update 🔄
APIRequestBuilder
ImportEventsView
MissionDataAndFlightSeqView


jbethanco:
Working perfect... maybe need to show something like, "No aircraft found" when nothing shows up, so the user knows that is the result, and it isn't still searching in the background or something... 
image

PBLogging/Network/PBSModels.swift
```diff
@@ -103,3 +103,19 @@ struct PBSCrew: Decodable {
...
+struct Aircraft: Decodable {
+    var result: [AircraftInfo]
+}
+
+struct AircraftInfo: Decodable, Hashable, Comparable {
+    var harmLocation: String = ""
+    var id: String = ""
+    var mds: String = ""
+    var tailNumber: String = ""
+    var unitCharged: String = ""
```
bmorris-revacomm, line 116: I think for our convenience, it would be nice to change tailNumber to serialNumber and have a custom coding key for it, so it doesn't crash when decoding. Ex:

private enum CodingKeys : String, CodingKey {
        case harmLocation, id, mds, serialNumber = "tailNumber", unitCharged
    }

PBLogging/Network/APIRequestBuilder.swift
```diff
@@ -14,6 +14,7 @@ enum APIRequestBuilder {
     case organization(organizationID: String)
     case organizations
     case pbsEvent(pbsID: String)
+    case tailNumbers
```
bmorris-revacomm, line 18: I would rename this to "aircrafts"

PBLogging/Network/APIRequestBuilder.swift
```diff
@@ -61,6 +62,12 @@ enum APIRequestBuilder {
             components.path = "/puckboard-api/v1/pbl/crewMember/\(pbsID)/events"
             print("BUILDER URL: \(components.url?.absoluteString ?? "nil")")
             return components.url!
+            
+        case .tailNumbers:
+            var components = standardComponents
+            components.path = "/puckboard-api/v1/tails"
+            print("BUILDER URL: \(components.url?.absoluteString ?? "nil")")
```
bmorris-revacomm, line 69: I would change "tails" to "aircrafts"

PBLogging/Network/APIRequestBuilder.swift
```diff
@@ -94,4 +101,5 @@ extension URLRequest {
     }
 
     static var organizations: URLRequest { APIRequestBuilder.organizations.urlRequest }
+    static var tails: URLRequest { APIRequestBuilder.tailNumbers.urlRequest }
 }
```
bmorris-revacomm, line 105: I would again rename to something like aircrafts

PBLogging/Network/ImportEventsView.swift
```diff
@@ -58,6 +58,18 @@ struct ImportEventsView: View {
...
+        print(#function)
+        network.getDataForRequest(.tails) { result in
+            switch result {
+            case .success(let data):
+                FileManager.default.saveToDisk("tails.json", data: data)
+            case.failure(let error):
+                print(error)
+            }
+        }
+    }
```
bmorris-revacomm, line 71: - since "tails.json" is a string, make sure to update if you change tails to a different word
- I would combine print(#function) with print(error). Since you're not printing anything if you hit the success case, it will just print the function name with no context

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -0,0 +1,108 @@
...
+    init() {
+        let aircraft: Aircraft = FileManager.default.decode("tails.json") ?? Aircraft(result: [])
+        let aircraftArray = aircraft.result
+        self.aircraftArray = aircraftArray
+        self.canSearchForAircraft = !aircraftArray.isEmpty
+    }
+    
+    @Published var canSearchForAircraft: Bool
+    @Published var aircraftArray: [AircraftInfo]
+}
```
bmorris-revacomm, line 108: - I think to be consistent with the rest of the project, it would be good to have the properties above the init
- I would rename "aircraft" to something like "decodedAircrafts" and then rename "aircraftArray" to just "aircrafts"
- also remember to update tails string here - I think it might be a good idea to put this json string in the Strings enum file that Sam made and then call it from there in both places

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+        }
+    }
+    
+    fileprivate func searchResults() -> [AircraftInfo] {
+        let results: [AircraftInfo] = aircraftPublisher.aircraftArray.filter({
+                                                                        $0.harmLocation.capitalized.contains(queryString.capitalized) ||
+                                                                        $0.tailNumber.hasPrefix(queryString) })
+            .sorted()
```
bmorris-revacomm, line 86: I would reformat the spacing here so it's easier to read:

let results: [AircraftInfo] = aircraftPublisher.aircraftArray.filter({
            $0.harmLocation.capitalized.contains(queryString.capitalized) ||
                $0.tailNumber.hasPrefix(queryString)
        })
        .sorted()

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -0,0 +1,108 @@
...
+                List(searchResults(), id: \.id) { item in
+                    Button {
+                        completion(item)
+                        isSearching = false
+                    } label: {
+                        HStack {
+                            Text(item.tailNumber)
+                                .font(.pblRegular(size: 14))
+                                .padding()
+                            Text(item.harmLocation)
```
bmorris-revacomm, line 67: I would put an HStack above the List that has texts for each column to improve the UX a little bit!

bmorris-revacomm (commented):
Some good work done here! Those completion handlers are pretty cool and make my brain dizzy 🤪 hahaha
I know there are a lot of comments, but it's mostly just easy, little stuff!

PBLogging/Network/APIRequestBuilder.swift
bmorris-revacomm: unless you need it to be tails for it to work with pbs? 🤔

PBLogging/Network/PBSModels.swift
```diff
@@ -103,3 +103,19 @@ struct PBSCrew: Decodable {
         "\(self.lastName ?? ""),\(self.firstName?.prefix(1) ?? "")"
     }
 }
+
+struct Aircraft: Decodable {
+    var result: [AircraftInfo]
+}
```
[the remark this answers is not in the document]
mcgarrys: I think "Aircraft" is plural 🤔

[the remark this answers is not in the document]
bmorris-revacomm: Oh my gosh, you're right hahaha. That's just as bad as filename being one word 😠 Well then maybe switch Aircraft and AircraftInfo, so the call site reads better for updateSortie? 🤔 Any thoughts? Am I overthinking it? haha

PBLogging/Network/APIRequestBuilder.swift
petermisik: tails is the api endpoint from PBS.

bmorris-revacomm: ahhh gotcha, then nevermind on this one haha 😊

bmorris-revacomm (approved):
Looks good! 🎉

PR #508, opened 2021-07-30
Generate swidtag file directly

jbethanco:
Discussion:
We were previously writing to the existing swidtag file in the resources folder and then copying it over.
New script writes the swidtag directly to the App Bundle in the Make Swidtag build phase.
We previously had the swidtag file in two locations in the development directory, both have been removed. 

Changes:
Correct spelling of Make Swidtag build phase
Update Make Swidtag build phase script to write file directly to app
Remove swidtag files


mcgarrys (approved):
Works here, swidtag file generates as expected to the app bundle ✅

PR #510, opened 2021-07-30
Change appearance.backgroundColor from UIColor(named: "background" to…

petermisik:
Discussion
Change appearance.backgroundColor from UIColor(named: "background") to .systemBackground

Add ✍️
None

Update 🔄
ContentView


jbethanco (approved):
Works.

PR #511, opened 2021-07-30
Fix/PBS-1586 and general updates

bmorris-revacomm:
- update optional date picker
- refactor sortie comments logic when generating pdfs - removes all unnecessary line breaks and adds line break between sortie comments
- update text and icons for buttons
- replace remarks on flight time view with sortie comment view
- refactor take off and land time UI
- refactor training events header
- remove BlurView


jbethanco (approved):
I dig it. Adds consistency and looks cleaner. 
🦸

PR #513, opened 2021-08-02
Fix/Training event tabs - (PBS-1587)

bmorris-revacomm:
- add logic to only show tabs of current crew positions


PBLogging/Views/SingleSortie/TrainingEvents/TrainingSwitchingView.swift
```diff
@@ -21,8 +21,12 @@ struct TrainingSwitchingView: View {
...
         VStack {
             Picker(selection: $crewPosition.animation(.linear), label: Text("Crew Position")) {
-                Text("Pilots").tag(CrewPosition.pilot)
-                Text("Loadmasters").tag(CrewPosition.loadmaster)
+                if sortie.crewLines.contains(where: { $0.isPilot }) {
+                    Text("Pilots").tag(CrewPosition.pilot)
+                }
+                if sortie.crewLines.contains(where: { $0.isLoadmaster }) {
+                    Text("Loadmasters").tag(CrewPosition.loadmaster)
+                }
```
jbethanco, line 29: Maybe could reduce the noise...
```
       
                if sortie.crewLines.contains { $0.isPilot }  {
                    Text("Pilots").tag(CrewPosition.pilot)
                }
                if sortie.crewLines.contains { $0.isLoadmaster }  {
                    Text("Loadmasters").tag(CrewPosition.loadmaster)
                }
     
```
 
If there is only pilots (or only loads, which would be impossible, only pilots might be impossible too?)... 
then there is a segmented control with only one segment, which looks kinda weird. But maybe just as weird as segments that present an empty screen. 
Also wondering if there is a way to load this for all different crew positions even if we don't have them implemented yet. Perhaps load an array with all the crew positions on the sortie, then a for each loop that makes the segments and loads the tags. Would require a loop initializing FlightAuthDutyCodes for every crewLine, then putting them into a set. Then looping the set for the segments. Let's tuple tomorrow, about it.

bmorris-revacomm: Oh, I didn't know you could remove where! haha
And yeah, I figured it would be impossible, but on the off chance they didn't add a pilot (or loadmaster) and clicked on it, I thought maybe it would be better to hide it? But maybe not?
Can we sync up today about this? I have some questions

PR #514, opened 2021-08-02
Fix/Aircrew header and flight card - (PBS-1600 and PBS-1601)

bmorris-revacomm:
- show text instead of aircrew list if there are no crew members
- only show flight card and training events card if sortie.crewlines is not empty
- rename remarks to comments


jbethanco (approved):
Fantastic.

PR #517, opened 2021-08-03
Fix/Remove CoreLocation - (PBS-1549)

bmorris-revacomm:
- remove all CoreLocation imports
- create Coordinate struct
- replace all instances of "CLLocationCoordinate2D" with "Coordinate"


jbethanco (approved):
Glad this was you and not me! 
Works great and no side effects.

PR #518, opened 2021-08-03
Fix/Textfield Auto Correct - (PBS-1599)

bmorris-revacomm:
- replace normal textfields with tabbed ones (where needed)
- refactor tabbed textfield
- fix textfields tab indices


PBLogging/Views/RecentSorties/RecentSortiesList.swift
```diff
@@ -26,16 +26,18 @@ struct RecentSortiesList: View {
...
                         
-                        TextField("Event Summary", text: $mission.summary, onEditingChanged: { _ in
-                            dataController.save()
-                        })
-                        .withBorderLabel("EVENT SUMMARY")
+                        TabbedTextField(text: $mission.summary, tabIndex: 2001, placeholder: "Event Summary")
+                            .onChange(of: mission.summary) { _ in
+                                dataController.save()
+                            }
+                            .withBorderLabel("EVENT SUMMARY")
```
jbethanco, line 40: Default behavior of TabbedTextField keeps the keyboard in uppercase. Doesn't seem right for the Mission name/ summary fields.

PBLogging/Views/Overview/PasscodeView.swift
```diff
@@ -32,17 +32,16 @@ struct PasscodeView: View {
             
             Group {
                 HStack {
-                    TextField("Unit name", text: $username)
-                    .onChange(of: username) { _ in checkPasscode() }
+                    TabbedTextField(text: $username, tabIndex: 1000, placeholder: "Unit name")
+                        .onChange(of: username) { _ in checkPasscode() }
```
jbethanco, line 36: This defaults to uppercase too.
The unit name and passcode fields no longer have padding and are tight together.

PBLogging/Views/SingleSortie/AircrewListView.swift
```diff
@@ -111,31 +111,26 @@ struct AircrewRow: View {
...
+                    .foregroundColor(crewLine.isAircraftCommander ? .pblForegroundPrimary : .pblForegroundSecondary)
+                    .opacity(determineImageOpacity())
+            }
+            .buttonStyle(PlainButtonStyle())
+            .disabled(!crewLine.isPilot)
+            
+            Group {
+                TabbedTextField(text: $crewLine.person.lastName, tabIndex: 5000, placeholder: "LAST NAME", autocapitalizationType: .words)
+                TabbedTextField(text: $crewLine.person.last4.onChange(enforce: .ssnFormatting), tabIndex: 5001, placeholder: "SSN", keyboardType: .numberPad)
+                TabbedTextField(text: $crewLine.flyingOrganization.onChange(enforce: .flyingOrgFormatting), tabIndex: 5002, placeholder: "0016", keyboardType: .numberPad)
```
jbethanco, line 128: Cool... didn't think it would work between rows, but the tabbing does.

jbethanco (changes requested):
Overall looks good. I think we should take a hard look at which fields should default to caps.

PBLogging/Views/RecentSorties/RecentSortiesList.swift
bmorris-revacomm: Oooooh good catch

jbethanco (approved):
Works great. Love that sentence vs word caps.

PR #519, opened 2021-08-03
PBS-1597 Add swidtag generation tests

mcgarrys:
[PBS-1597](https://jira.il2.dso.mil/browse/PBS-1597)

### Discussion
This PR adds tests ensuring the swidtag file generates successfully after the "Make Swidtag" build script is run, and that the content of the file is as we expect it to be. This means that if anything is different between the string we get and the one we expect (whether it be the version number, build number, or even if someone changes a word accidentally) a test should fail. If the file stops existing for some reason, a test should also fail.

### Testing
Navigate to the "Make Swidtag" build script (Project > Build Phases > Make Swidtag) ➡️ 
- [x] Change something in the swidtag string on line 10, and verify this breaks the **testSwidtagHasExpectedContent** test ❌


jbethanco (approved):
Works great... tried to pass the test with various mods... could not.

PR #523, opened 2021-08-04
Fix/pbs 1604 search by prefix

petermisik:
# Discussion  💬
Current solution searches the harm location and sees if the string contains queryString.  This gives us a more granular control to provide the user the result based on the prefix of each of the words in the harm location.

# What this PR Does 🤔 
Add ability to search by Harm Location using prefix.  Gives us a generic way to search prefix on long strings and provide a boolean result.  This search ability can be used in other areas.  Add highlights to headers to indicate which field they're searching.

# Resolves 
[PBS-1604](https://jira.il2.dso.mil/browse/PBS-1604?filter=-1) - Update tails search to search by word prefix.

# Testing 🔬 
- [X] - Type the letter "C" in the field - Returns Charleston, Charlotte, and CA
- [X] - Type in "01" in the field - Returns aircraft with serial numbers beginning with 01
- [X] - Type in "UNK" - Return "No Aircraft Found"


jbethanco:
I'm not sure if highlighting the column header feels intuitive. Thinking we could bold the text of the found field. e.g. if 01-8031 is found with 01, then serial number 01-8031 is bolded. If zip code 01203 is also found with the same search, then the harm location is also bolded. Not the titles, but just the text. 

The method of highlight feels disabling to me. 
image

PBLogging/Extensions/String+Ext.swift
```diff
@@ -185,4 +185,10 @@ extension String {
         let attributedText = NSAttributedString(string: self, attributes: textAttributes)
         return attributedText.size()
     }
+    
+    func doesComponentHavePrefix(queryString: String) -> Bool {
+        let characterSet = CharacterSet(charactersIn: " ,-")
```
jbethanco, line 190: I don't think the comma is necessary here.

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -114,12 +120,39 @@ extension View {
...
+        } else {
+            if serialNumberCharacters.isSuperset(of: queryString) {
+                self.isSearchingBySerialNumber = true
+                self.isSearchingByHarmLocation = false
+            } else {
+                self.isSearchingBySerialNumber = false
+                self.isSearchingByHarmLocation = true
+            }
+        }
+    }
```
jbethanco, line 149: Instead of highlighting the column header, I think we should be highlighting the found value in the rows. Could probably make a function that returns true if the passed in item text has a component with the prefix matching the search string. That could be used for setting the font to bold on the one item, but could also be used by the search filter.

jbethanco (changes requested):

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -95,14 +95,12 @@ fileprivate struct AircraftSearchResultsView: View {
...
-        })
-        .sorted()
-        
-        return results
+    func fontForSearchResult(item: String, queryString: String) -> Font {
+        if item.isSerialNumberSet(queryString: queryString) {
+            return !queryString.isEmpty && item.hasPrefix(queryString) ? .pblBold(size: 14) : .pblRegular(size: 14)
+        } else {
+            return !queryString.isEmpty && item.doesComponentHavePrefix(queryString: queryString) ? .pblBold(size: 14) : .pblRegular(size: 14)
+        }
```
bmorris-revacomm, line 103: Again, I feel like "item" isn't descriptive enough, but I also just don't like using the word "item" for anything haha. Maybe there is a better word or better way to name the function?
Maybe something like:
fontFor(_ result: String, using queryString: String)

PBLogging/Extensions/String+Ext.swift
```diff
@@ -185,4 +185,16 @@ extension String {
         let attributedText = NSAttributedString(string: self, attributes: textAttributes)
         return attributedText.size()
     }
+    
+    func doesComponentHavePrefix(queryString: String) -> Bool {
+        let characterSet = CharacterSet(charactersIn: " -")
```
bmorris-revacomm, line 190: Thinking maybe "componentHasPrefix" might be a better name since the built-in function is "hasPrefix" - seems more consistent. I would also put _ before queryString, so it's not left in the parameter when called.

PBLogging/Extensions/String+Ext.swift
```diff
@@ -185,4 +185,16 @@ extension String {
...
+        let uppercasedComponents = self.uppercased().components(separatedBy: characterSet)
+        return uppercasedComponents.contains { $0.hasPrefix(queryString.uppercased()) }
+    }
+    
+    func isSerialNumberSet(queryString: String) -> Bool {
+        guard !self.isEmpty else { return false }
+        let numberSet = Set("0123456789-")
+        return Set(self).isSubset(of: numberSet)
+    }
```
bmorris-revacomm, line 199: - queryString is unused here
- I think this could use a better name? I feel like I need to look at what the function is doing to understand why this is being called. I think the word "set" threw me off? 🤔

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -60,31 +60,31 @@ fileprivate struct AircraftSearchResultsView: View {
...
+                                .font(fontForSearchResult(item: item.serialNumber, queryString: queryString))
+                                .padding(.leading, 10)
+                                .frame(width: 95, alignment: .leading)
                             Text(item.harmLocation)
-                                .font(.pblRegular(size: 14))
+                                .font(fontForSearchResult(item: item.harmLocation, queryString: queryString))
                                 .frame(maxWidth: .infinity, alignment: .leading)
                             Text(item.unitCharged)
-                                .font(.pblRegular(size: 14))
                                 .frame(maxWidth: 200, alignment: .leading)
```
bmorris-revacomm, line 88: this font needs to stay

PBLogging/Modifiers/AircraftSearchButtonModifier.swift
```diff
@@ -114,12 +112,19 @@ extension View {
...
         self.canSearchForAircraft = !aircraftArray.isEmpty
     }
+    
+    func searchResults(queryString: String) -> [AircraftInfo] {
+        self.aircraftArray.filter {
+            $0.harmLocation.doesComponentHavePrefix(queryString: queryString) ||
+            $0.serialNumber.hasPrefix(queryString)
+        }
+        .sorted()
+    }
```
bmorris-revacomm, line 129: I would rename slightly to searchResults(of queryString: String)

bmorris-revacomm (changes requested):
just some small changes!

bmorris-revacomm (approved):

jbethanco (approved):

PR #524, opened 2021-08-04
Rename and Add New NowSecure Token

jbethanco:
Updated Now Secure Token in Github secrets and added a wiki page to describe how it is done.
https://github.com/airmencoders/Logging-iOS/wiki/How-to-update-the-NowSecure-Token

This updates the NowSecure workflow script with new name.
New name makes it easier to understand what it does and where it came from. 

To test:
- [x] Make a branch off of this one, and then push it and try to run the NowSecure action.  Verify that [NowSecure](https://lab.nowsecure.com/) has received the app and that it does a full test.


jbethanco:
Seems the second time, the test ran fully.

mcgarrys (approved):
Just reran the test branch on NowSecure, looks good 🔥

PR #525, opened 2021-08-04
PBS-1582 Add back edit aircrew button

mcgarrys:
**Discussion** 🗣

This PR adds back the "Edit" button to the "Edit Aircrew" section, bringing back the ability for users to delete aircrew with the minus buttons on the side of the list. The ability to delete by swiping left on a row was a bit unclear for users new to iOS.

**Important Notes** ✍️
- The column titles will change their position to match the columns change when in edit mode. 
- The column titles and the textfields within the aircrew rows have animations applied to them, so as to match up with the animation that occurs when the list enters/exits edit mode.


bmorris-revacomm (commented):
It looks good, except there's a weird little bug 🪲
If you are in edit mode and select someone as ACC (except for the very first person), and then delete them, it moves that person to the top and then deletes a different person.

Additionally, I think it would be nice to disable the edit button when there are no crew members in the list :)

bmorris-revacomm (approved):
Looks good! 🐠

PR #527, opened 2021-08-06
Increase Detail of Sample Mission for JSON output/testing

jbethanco:
## 💬 Discussion 
In order to provide future intended output to scheduling, this PR includes updates to a sample mission to include information not yet collected in the app including pilot initials and isReadyForReview..
 
## 🤔 What does it do? 
Update sample Mosher mission to include:
flight times, conditions, training events, sortie type, and initials, and isReadyForReview
Add Comparable extension to Flight struct as missions weren't sorting by date on generated forms.

Add logic in 781 pdf generation to sort the flights and add initials as needed to forms.
(If there is the same initials on all forms, then the initial is only included in the A block on the 781, if any initials are different, then A through F blocks receive the per sortie initials.)

## 🧾 Checklist
- [X] Unit and UI tests pass locally with my changes
- [X] Unit and UI tests pass remotely with my changes
- [X] I have added tests that prove my fix is effective or that my feature works

## 🔬To test
- [x] Sorties are sorted by date from oldest to newest on generated Form 781s.
- [x] Initials show up on the Operation Scorpion mission 781 pdf.


mcgarrys (approved):
Looks good 👽. Sorties appear in the correct order on generated 781, and initials are showing up on the pdf (in sim and on device). All tests pass locally for me.

I did notice that the Mission Sortie Info forms are sorted by most recent takeoff time, rather than the oldest, is that something that matters on the form?

jbethanco:
> sion Sortie Info forms are sorted by most recent takeoff time, rather than the oldest, is that something that matters on the form?
Not sure if the order of those matter. Will have to check.

PR #528, opened 2021-08-06
Replace print with new functions

jbethanco:
### Discussion
This PR replaces the print function with seven new functions:
printInfo / printShortInfo
printWarning / printShortWarning
printError / printShortError
and printLine
standard print is marked unavailable to use

The long form functions take advantage of a parameter defaults so that file, function, and line parameters are taken from the call site of the print, without having to type them in.

The result is a print statement that provides call site location context by default without extra work from the user.  

In addition to that, each has its own emoji prefix for easier identification of different types of prints which also makes it easy to differentiate our comments from system comments. 

The last and maybe most important benefit, is that it keeps our print statements out of release builds and allows the compiler to completely optimize out the print statements. This happens because the bodies of all the new print functions are wrapped with '#if DEBUG' compilation conditionals. In release builds they are empty and the function calls are optimized away as if they do not exist.  

This:
 

Results in:
 

### Changes:
Adds Print.swift file with functions.
Replaces all existing print calls.


bmorris-revacomm:
What about instead of having to use printLine, you just add an empty line to the end of all the other print statements? That should make the console easier to read anyway 🤔

jbethanco:
> What about instead of having to use printLine, you just add an empty line to the end of all the other print statements? That should make the console easier to read anyway 🤔

Want it to be explicit and similar to the regular print. Sometimes you will want row after row, together for comparison.

bmorris-revacomm:
> > What about instead of having to use printLine, you just add an empty line to the end of all the other print statements? That should make the console easier to read anyway 🤔
> 
> Want it to be explicit and similar to the regular print. Sometimes you will want row after row, together for comparison.

Ahhhh okay, makes sense

bmorris-revacomm:
Do we need a warning one? I'm just wondering when we should use warning over error 🤔 It looks like you are only using it twice right now, but I feel like those could easily be replaced with error

jbethanco:
> Do we need a warning one? I'm just wondering when we should use warning over error 🤔 It looks like you are only using it twice right now, but I feel like those could easily be replaced with error

Was thinking about places that aren't actual errors, but might have suboptimal side effects. Like not having the downloaded file for searching. So you'll know why something isn't working.

PBLogging/ViewModels/Auditing/OtherTimeAuditor.swift
```diff
@@ -43,7 +43,7 @@ struct OtherTimeAuditor: Auditing {
     }
     
     mutating func calculateTime() {
-        print(#function)
+        printInfo(#function)
```
mcgarrys, line 46: Should this be **printInfo("")**? It seems #function ends up in the print statement anyway.

mcgarrys (approved):
Works great here. Left a comment about a super tiny call site thing.

I'm here for the emojis in the debug console 🙌

PR #529, opened 2021-08-06
Add button modifier

bmorris-revacomm:
- create button modifier for disabling and reducing opacity
- implement where needed
- move Coordinated to more appropriate folder


jbethanco (approved):
Works good here.

PR #530, opened 2021-08-06
PBS-1639 Fix textfield expansion bug

mcgarrys:
### Discussion 💬
This PR fixes the dreaded textfield expansion bug that occurred when typing in our custom **TabbedTextField**. When a user typed text that was longer than the TextField itself, the Textfield would expand to show the entire string, and expand the entire view with it. 

The fix ended up being as simple as setting the property **setContentCompressionResistancePriority(.sceneSizeStayPut, for: .horizontal)** for the UITextField being used, and adding some **.leading** and **.trailing** padding to the **TabbedTextField** view.

### Changes 🤔
- Fix textfield expansion bug
- Add 50 character limit format to **StringFormat** and apply it to all Mission Data and Fuel/Airland/Airdrop textfields. This was the number agreed upon (for now) in the most recent dev/design sync.
- Remove "ing" from names of string formats. E.g. change **.icaoFormatting** to **.icaoFormat**.

### Resolves 
[PBS-1639](https://jira.il2.dso.mil/browse/PBS-1639) - Fix textfield expansion bug

### Testing Items 🔬

- [ ] In every Mission Data and Fuel/Airland/Airdrop textfield, type as much as you can, and ensure you hit the 50 character limit. 
- [ ] Ensure that typing a string longer than the textfield itself does not result in the view expanding.

### Resources 📚
- [setContentCompressionResistancePriority(_:for:)](https://developer.apple.com/documentation/uikit/uiview/1622526-setcontentcompressionresistancep)
- [UILayoutPriority](https://developer.apple.com/documentation/uikit/uilayoutpriority)


PBLogging/Extensions/String+Ext.swift
```diff
@@ -84,6 +84,10 @@ extension String {
         }
     }
     
+    func enforceCharacterLimitFormat() -> String {
+        String(self.suffix(50))
+    }
+
```
jbethanco, line 90: I'm not sure if suffix gives the desired behavior vs prefix. With suffix, it give the user the experience that they are just still typing even though the 50 characters to the left, which may not be visible, are getting trimmed off the front. With prefix, they will get immediate feedback that they can not type any more and have reached a hard limit.

jbethanco (approved):
Everything works regarding expansion.

PR #531, opened 2021-08-09
Fix/items for now secure

jbethanco:
### Discussion: 
Resolves remaining issues in NowSecure and brings score to 100.

### Changes:
Remove device and device version from User Agent String
Change passcode textfield to secure text field
Update group ID in NowSecure build script to the current Tron ID

### To Test:
- [ ] Does the passcode screen still work?


bmorris-revacomm (approved):
works  🧆

PR #533, opened 2021-08-09
Fix/781 pdf generation

bmorris-revacomm:
- fix flight auth duty code format so aircrew is sorted correctly
- rename file and mission to be more specific


jbethanco (approved):
Nice. Thanks for this.

PR #534, opened 2021-08-09
Add character limits to all textfields

mcgarrys:
### Discussion 💬
This PR adds and updates character limits to all textfields throughout the app, based on real context regarding the 781 and 791 data. The majority of fields will have the 50 character limit as discussed in the last dev/design sync, with the only exceptions being fields that we know will never exceed a certain amount of characters, or fields we know may need more than 50.

### Changes 🤔
- **enforceCharacterLimit** function and **.characterLimit** enum case now take a specified character limit as a parameter
- **enforceAndUpdate** functions can take a character limit as a parameter as well.
- One version of **enforceAndUpdate** was using generics to be able to take any kind of Int as a parameter, but I removed this functionality as we are only using it for Int16s.
- Tests have been updated to reflect the above changes.

### Resolves ✅
[PBS-1662](https://jira.il2.dso.mil/browse/PBS-1662) - Add character limits to textfields

### Testing Items 🔬
Ensure that the character limit works with the correct specified limit for the following textfields:

**Recent Sorties**
- [x] Event name - **50**
- [x] Event summary - **100**

**Edit Aircrew**
- [x] Last name - **50**
- [x] Flight auth duty code - **10**
- [x] Reserve status - **2**

**Mission Data**
- [x] All fields (excluding Takeoff ICAO and Land ICAO) - **50**

**Fuel/Airland/Airdrop**
- [x] All fields - **6**

**Flight Time**
- [x] All fields - **4**

**Tanker Data**
- [x] All fields - **50**

**Issues**
- [x] Aircraft command - **6**
- [x] Aircraft type - **8**
- [x] Aircraft number - **10**
- [x] Aircraft call sign - **10**
- [x] Aircraft code and home station - **30**


PBLogging/Views/SingleSortie/AircrewListView.swift
```diff
@@ -135,12 +135,12 @@ struct AircrewRow: View {
             .disabled(!crewLine.isPilot)
             
             Group {
-                TabbedTextField(text: $crewLine.person.lastName, tabIndex: 5000, placeholder: "LAST NAME", autocapitalizationType: .words)
+                TabbedTextField(text: $crewLine.person.lastName.onChange(enforce: .characterLimit(limit: 50)), tabIndex: 5000, placeholder: "LAST NAME", autocapitalizationType: .words)
```
jbethanco, line 138: Tiny thing. Whenever I find a repeating word in a function call, I try to get rid of it so long as the context is still understandable. Omitting the argument label here could do that, and in the modifier... 
e.g. .characterLimit(40) 
I'm just glad it wasn't .characterLimit(limit: limit)

jbethanco (approved):
Works good. I verified each limit. I was worried this would get ugly, but that is a pretty clean solution. 
I left a comment about omitting arguments.

PR #535, opened 2021-08-09
Update App Icon set with new images

jbethanco:
Replaced all images in the app icon set

### Discussion:
Old image had borders with white background and some aliasing. This fixes that. 

### To test:
- [x] If you delete the app from your device and do a clean build, does the app get an icon?


mcgarrys (approved):
Works here ✅

PR #537, opened 2021-08-10
PBS-1667 Fix flight time padding

mcgarrys:
**Discussion**
This PR fixes a small padding issue in the Flight Time section. The sortie comments view had too much padding, and was not lined up with the rest of the view.


bmorris-revacomm (approved):
looks good!

PR #538, opened 2021-08-10
Fix/Copy AC to new sortie - (PBS-1660)

bmorris-revacomm:
- set aircraft commander when creating new sortie


mcgarrys (approved):
Works great 💥

PR #539, opened 2021-08-10
PBS-1669 Add UI test for passcode screen

mcgarrys:
### Discussion 💬
This PR adds a UI test to verify that the passcode screen is working. 

### What this test does 🤔
- Clears the **.usernameForPasswordCheck** key in UserDefaults using a new launch argument in the SceneDelegate (if it is not empty, the app will log itself in).
- Enters a test unit name, which in this case is "UITesting".
- Enters a passcode we know is incorrect, and asserts that the "Missions" navigation title is not in the scope of the view.
- Enters a passcode we know is correct, and asserts that the "Missions" navigation title is now in the view (meaning the app has logged in and navigated to the missions overview screen).

### Resolves ✅
[PBS-1669](https://jira.il2.dso.mil/browse/PBS-1669) - Add passcode screen UI test

### Testing Items 🔬

- [ ] Ensure that **testPasscodeEntry** passes locally


bmorris-revacomm:
I'm so confused, please tell me where the string for correctPasscode is coming from because I can't find it haha

bmorris-revacomm (approved):
looks good to me! 🎃

PR #540, opened 2021-08-10
Fix/Training event name truncation - (PBS-1661)

bmorris-revacomm:
- update font and font sizes for training event names and IDs
- wrap text to prevent truncation

Note:
Currently, only one is needing the text wrap (AP23YM) - making the font smaller fixed the rest. BUT, now the divider after this one is off from the divider in the columns. I don't think there is a way to fix the dividers lining up, without things getting really messy, and I don't think it's a huge deal, especially if we plan to redo these views in the near-ish future.


jbethanco (approved):
Works for me. 🦧
I don't think there would be any problem using ILS and removing the long form to make it fit.

PR #541, opened 2021-08-11
Feature/Add first name to UI

bmorris-revacomm:
- add first name to both aircrew views
- adjust frames and padding as needed to prevent truncation, ugly UI, etc


jbethanco (approved):
Works well here. 🧳

PR #542, opened 2021-08-12
Refactor Event Import / Add modifiers

jbethanco:
### 💬 Discussion:

This PR is in preparation for uploading to PBS. 

#### New Select Event View 

A mission uploaded to PBS, must have been originally downloaded from a PBS event to have its ID.

If a user created a mission locally, they can not upload it. In order to allow them to upload it, we need the capability to let them associate it with an existing PBS event so that it can obtain it's ID (UUID). 

The existing ImportEventsView gives them the ability to import an event as a new mission, but not associate it with a current mission. We need that view to be reusable for both download and for association. 

This PR introduces **getPBSEventSheet** in order to do that. It is a sheet view modifier that presents a sheet with a custom title that allows the user to login and select an event from PBS. It then returns that event in a closure. 

Example:

```swift
YourView()
.getPBSEventSheet(isPresented: $isImportSheetPresented, title: "Select Mission to Download", networkClient: network) { pbsEvent in
    self.addEventFromPBS(pbsEvent)
}
```

Network Client is injected here for testability and to be explicit, as sheets do not capture SwiftUI environment variables. There is no way to forget to add it as it is a parameter here. You also can not initialize a view model with an environment variable. 
 
#### ⌨️ Login

The current login view is only used by the importEventsView. The login needs to be reusable for other interactions with PBS, like upload, or any other future work. 

This PR introduces **ShowIfLoggedIn**, a view modifier

Example:

```swift
YourView()
   .showOnlyIfLoggedInToPBS()
```

If they are logged in, it shows YourView(). Otherwise it shows the LoginWebView until they login. 

Currently there is no way to simulate a network connection so that you can preview views that require network data. This makes designing the views very difficult as you had to run the simulator and login to platform 1 to view them. 

This PR introduces **MockNetworkClient**.

MockNetworkClient can be injected into canvas previews allowing their view models to "download" fake JSON. 

### 🎯 Specific Actions:

Fix LoginWebView so that it doesn't reload when you exit and return to the app to get your OTP.

Add arrowRight to SFSymbols

Downloaded event crew members are only added if they have a first and last name

Update sample data for new PBS structs

Add MockNetworkClient and NetworkClientConforming protocol to make use of it
Add sample JSON file used by MockNetworkClient for use in previews

Add ShowIfLoggedInModifier
- If not logged into P1, shows the P1 login. 
- Otherwise, shows your content

Move LoginState enum to LoginWebView in support of the new modifier

Add endpoint to APIRequestBuilder for getting events by organization ID

Add SelectPBSEventView, GetPBSEventSheetModifier, GetPBSEventViewModel
- End result is a sheet view modifier we can use anywhere in the app to select an event from PBS. It returns the user selected PBSEvent.

Delete ImportEventsView
Delete PBSModels file
Add PBSStructs file
Add PBSStructsTests

Update PBSEventCard to show takeoff and landing ICAO
Change PBSEventCard to show download arrow icon instead of right chevron
Add padding to same

Refactor MissionsOverview to use getPBSEventSheet modifier instead of ImportEventsView

Add PBSEventCard canvas previews
Add SelectPBSEventView canvas previews

### 🧾 Notes:

The SelectEventView does not look polished.  Wanting to the get this merged in, then pass off the UI tweaking to someone else since the canvas is working. 

If you want a live walkthrough, let me know. 

### 🔬 To Test:

Not everyone has IL4 access. If you want to pair and takeover my computer to test, let me know. 

- [ ] Can you login and import missions.
- [ ] Are newest missions on top?
- [ ] Persons without a first or last name on a PBS event not added to imported missions.
- [ ] If you are already logged in, does the login screen not appear.
- [ ] Canvas previews work for SelectPBSEventView and PBSEventCard


bmorris-revacomm (approved):
looks good from what i can tell 👺

PR #545, opened 2021-08-16
PBS-1683 Implement "isSim" on mission

mcgarrys:
### Discussion 💬
This PR adds the ability to toggle whether or not the mission is a sim. Logic is also added to handle a case where the user has logged training events and wants to switch the mission to be a sim. The user will be alerted that the training events for the mission will be deleted, and be given a destructive option to continue, or a secondary option to cancel the action. 

Please let me know what you think of the alert title and message, as well as the name for the destructive button on the alert (which is currently just "Continue").

### UI Changes 👁
IMB_0ofigs

### Resolves ✅
[PBS-1683](https://jira.il2.dso.mil/browse/PBS-1683) - Implement "isSim" on mission

### Testing Items 🔬
Set up: Navigate to the training events section for any mission's sortie, and add several training events if there aren't any. Repeat this with at least one other sortie in the mission. 
- [x] Navigate to the sorties list for the mission, and tap the "Simulation" toggle. Ensure the toggle is turned on, and an alert is presented on the screen.
- [x] Tap the "Cancel" button on the alert. Ensure the alert disappears, and the toggle is turned back off.
- [x] Tap the "Simulation" toggle again, and this time tap on the "Continue" button (this button has a destruction action, which is why it is red). Ensure the "Simulation" toggle remains on.
- [x] Navigate to the mission data view for every sortie in the mission. Ensure all training events have been deleted from all sorties. 
- [x] Finally, navigate to the home screen, close the current app session, and re-launch the app. Ensure the changes you made to the mission persist, and the "Simulation" toggle is on.


bmorris-revacomm:
can you make the toggle our blue color instead of green? 🤔

PBLogging/Views/RecentSorties/RecentSortiesList.swift
```diff
@@ -127,6 +124,33 @@ struct RecentSortiesList: View {
...
+                                deleteMissionEventRecords()
+                            },
+                            secondaryButton: .cancel {
+                                withAnimation {
+                                    mission.isSim = false
+                                }
+                            }
+                        ))
+                    }
+                }
```
jbethanco, line 149: I think the function to show the alert should probably be extracted out. 
In other cases where we are looking for a specific quality of a Mission or Sortie where we are looping, we have abstracted it away to a property on the Mission or Sortie in the accessors file. I think that would work well here. e.g.
`mission.hasTrainingEventsLogged` 
This way we can avoid having a loop called in a view. 

Also, this check seems to only check for training events in the first sortie. I don't think it would trigger if there are training events in the second or third, but not in the first. 

This would also trigger the alert for each crewLine with training events. Due to how the AlertProvider works, you won't notice it, but we should probably break on the first one

PBLogging/Views/RecentSorties/RecentSortiesList.swift
```diff
@@ -82,16 +79,16 @@ struct RecentSortiesList: View {
         .navigationBarTitle(Text(mission.name))
         .navigationBarItems(trailing:
                                 HStack {
-                                    uploadButton
+                                    uploadMissionButton
                                     addSortieButton
                                         .accessibility(identifier: "addSortieButton")
-                                    editButton
+                                    editSortiesButton
                                     InfoButton()
```
jbethanco, line 86: Nice.

PBLogging/Views/RecentSorties/RecentSortiesList.swift
```diff
@@ -127,6 +124,33 @@ struct RecentSortiesList: View {
         })
     }
     
+    var isSimToggleButton: some View {
```
jbethanco, line 127: This is a Toggle. Perhaps just "simToggle"

jbethanco (commented):
Looks good. A couple of comments include. 
I believe the functionality should work in either direction. Switching from sim, or switching to sim. If they have logged events in either mode, they should get the warning and deletion upon the switch.

jbethanco (approved):
Passes all tests.

PR #547, opened 2021-08-18
Refactor Binding onChange extension

jbethanco:
### Discussion: 🥏
Previous version was untestable because bindings are currently untestable. 
This brings the logic out of the binding and into the call site
Call sites read near identical to previous version

### Changes: 🥩
Refactors onChange Binding extension
Updates all call sites
Adds doc comment to onChange
Deletes StringFormat enum

### Check: 👀
- [x] Character limits are all the same

### Test: 🔬
- [x] ICAOs, SSAN, crew Flying Org, and character limits all behave as in previous versions


mcgarrys (approved):
Looks good 📲  💫. All formats and limits work as expected.

PR #549, opened 2021-08-18
PBS-1687 Add Sortie+TrainingEvents tests

mcgarrys:
### Discussion 💬
In short, this PR adds a test to make sure the **getCrewMembersAndTrainingEvents()** extension function does what it is supposed to do. In this case, that means identifying how many unique events are in a Sortie, identifying which crew members have those events logged, as well as how many times they logged that event.
### What this test does 🤔
- Assigns mission event records (some with the same ID to test the function returns several crew members for one event) to several different crewlines.
- Assigns these crewlines to the sortie, and calls **getCrewMembersAndTrainingEvents()**.
- Asserts the events returns have the expected eventIDs.
- Asserts that each returned event has the expected crew member(s) by checking the **lastName** property.
- Asserts that the **totalAccomplished** for each event is equal to the sum of its crew member's **numAccomplished** properties.
### Notes ✍️
- [PBS-1693](https://jira.il2.dso.mil/browse/PBS-1693) will rename the function to be called **getTrainingEventsAndTotals()**. The **CrewMember** struct will get some new properties as well. I took a look at the updated code that Beth wrote, and from what I can tell, it shouldn't affect what this test is doing. 
### Resolves ✅
[PBS-1687](https://jira.il2.dso.mil/browse/PBS-1687) - Unit Test - Sortie+TrainingEvents
### Testing Items 🔬
- [ ] Do the tests check what the function is actually doing, and run?


bmorris-revacomm (approved):
looks good to me!

PR #556, opened 2021-08-20
PBS-1598 Add JSON generation tests

mcgarrys:
This PR adds a unit test that verifies our JSON output for missions is correct. Currently, we are testing JSON by counting individual characters because arrays with randomly ordered elements make it difficult to do a direct comparison of data. 

- Loads the "all fields filled" mission from the sample data.
- Creates and adds several mission event records to the mission.
- Assigns a static UUID to all ids, within the mission, as we cannot test for random UUID generation.
- Generates JSON, and converts it into a dictionary counting all unique characters in the string. 
- Asserts that all character counts are equal when compared to the expected output dictionary.

[PBS-1598](https://jira.il2.dso.mil/browse/PBS-1598) - Test JSON Generation

- [x] Does this test, test what it is supposed to test well?
- [x] Does the test run locally?

### Notes :writing_hand:
- The function used to generate character count dictionaries for strings has been added as a String extension, so it can be used for this test, as well as in the Mission+EncodableTests.


PBLogging/Extensions/String+Ext.swift
```diff
@@ -190,4 +190,17 @@ extension String {
         let uppercasedComponents = self.uppercased().components(separatedBy: characterSet)
         return uppercasedComponents.contains { $0.hasPrefix(queryString.uppercased()) }
     }
+
```
jbethanco, line 193: Should we include it in the project if it is only being used by tests. We could have a String+Ext in the test support folder?

jbethanco (approved):
Works good.

PBLogging/Extensions/String+Ext.swift
mcgarrys: Good point, String+Ext in the test support folder sounds like a good idea. I wonder if there are any other extensions we are only using for tests 🤔

PR #557, opened 2021-08-20
PBS-1602 Add AircraftPublisher tests

mcgarrys:
This PR adds units tests that verify the **AircraftPublisher** _ObservableObject_ functions correctly. The tests cover everything within the bounds of the current search functionality (as far as I can tell). I also tried out a version of the Gherkins format.

1. Asserts that **AircraftPublisher** initializer works correctly.
2. Asserts that we get expected search results for the following queries:
- Empty string
- A partial serial number (in this case, one that would return results with different serial numbers).
- A specific serial number.
- A partial/incomplete location keyword (in this case, a single character to verify results with different harm locations are returned).
- A specific location keyword.
3. Asserts that when more than one result is returned, the aircraft are sorted by serial number.

[PBS-1602](https://jira.il2.dso.mil/browse/PBS-1602) - Write tests for Aircraft Publisher used by Aircraft

- [x] Do the tests test what they are supposed to test well?
- [x] Do the tests run locally?

### Resources :books: 
[5 ways to make scenarios interesting to read: Given-When-Then best practices](https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/) - Got a few ideas from this.


LoggingTests/AircraftPublisherTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        
+        XCTAssertEqual(locationSearchResult.count, 3)
+        
+        // And anytime we search the aircraft array with a query that produces more than one result...
+        
+        let moreThanOneSearchResults = aircraftPublisher.searchResults(of: "")
+        
+        // Then those results should be sorted by serial number.
+        
+        XCTAssertTrue(moreThanOneSearchResults[0] < moreThanOneSearchResults[1])
```
jbethanco, line 87: This could maybe randomly be correct if something changes as it is only comparing two things. Perhaps...
let sorted = moreThanOneSearchResults.sorted { $0.serialNumber < $1.serialNumber }
        
        XCTAssertEqual(sorted, moreThanOneSearchResults)

LoggingTests/AircraftPublisherTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        XCTAssertFalse(aircraftPublisher.canSearchForAircraft)
+    }
+
+    func testSearchResults() throws {
+        
+        // Given we have an array of mock Aircraft data...
+        
+        let data = SampleJSON.aircraftSampleData.data(using: .utf8)!
+        FileManager.default.saveToDisk("testJSON.json", data: data)
+        let aircraft: Aircraft = FileManager.default.decode("testJSON.json") ?? Aircraft(result: [])
```
jbethanco, line 32: Might be good to delete the file after the test, but probably doesn't harm anything.

jbethanco (approved):
Works good. 
Couple of minor comments.

jbethanco (approved):
🤘

PR #560, opened 2021-08-21
Add pilot review initials

jbethanco:
Adds text field to add pilot review initials. 

Adds a text field to put in pilot initials. If they tap into the field and it is empty, and there is a designated aircraft commander with a first and last name, then it puts their initials. 

[PBS-1730](https://jira.il2.dso.mil/browse/PBS-1730)  

- [ ] If you have a designated Aircraft Commander and you tap into an empty initials field, does it add the initials? 

### UI Changes :eye: 
Adds initials at bottom right... 
image


bmorris-revacomm (approved):
Looks good! Do we want it to be a tabbed textfield though?

jbethanco:
> Looks good! Do we want it to be a tabbed textfield though?

Wanted to, but needed the "onEditingChanged:" to do the auto initial thing. Maybe we could look on adding that as an optional parameter to our TabbedTextField.

PR #561, opened 2021-08-22
Fix redraw bug when upload mission view shown

jbethanco:
This fixes a UI glitch.
A sheet which shares the same observable object as the view that presents itself causes problems. 

**Long winded example of the problem:**
CarView presents a sheet with ChangeCarColorView.
CarView and ChangeCarColorView both share an Observable Object called "Car"
CarView has a Text showing the car's color. 
You then present the ChangeCarColorView so that you can change the color. 
By changing the car's color on ChangeCarColorView, CarView needs to redraw, because it is showing the cars color. Since the sheet showing ChangeCarColorView is attached to the CarView, CarView's required redraw causes ChangeCarColorView to redraw as well. 

**The solution** is to hide anything Car related on CarView while the ChangeCarColorView is presented in the sheet. This allows changes to Car without causing CarView to redraw. 

Add large comment concerning fix
Change mission in Upload Mission View to observable object
Remove state boolean that was used to track ready to upload
Uppercase SARM
 
#### Before:
before

#### After:
after


bmorris-revacomm:
maybe nitpicking, but I think the done button would be better with either a check mark or no icon at all 🤔 Currently, the icon would make me think it was a cancel action and would wipe/not save anything I just did

jbethanco:
> maybe nitpicking, but I think the done button would be better with either a check mark or no icon at all 🤔 Currently, the icon would make me think it was a cancel action and would wipe/not save anything I just did

Not part of this PR. Planning to send the UI back to Rachael now that we have the workflow worked out.

PR #563, opened 2021-08-24
Fix/remove scene delegate dependencies

jbethanco:
In order to make the code injectable/testable removes dependency on scene delegate to get at the view context. Part of which was changing how we add sample data. 
It is no longer added by shaking, but rather long pressing the logo for 4 seconds.
This also helps clear the path to switching to the SwiftUI App Life Cycle. 

An unrelate test for AircraftPublisher was updated during this as it was discovered that if the simulator has the actual JSON file downloaded, it causes the test to fail. 

Remove scene delegate as means to get the view context
and instead pass it in as needed in the following locations:
- All PDF generators
- All PDF generator tests

Replace "Shake to Add Missions" in debug mode
with "Long Press Logo for 4 seconds" to add missions in any mode

Add "loadSampleMissions" function to SampleData

Clean up PDFPreviewView initializer

- [x] Verify PDF generation still works
- [x] Verify long press of logo adds sample data

[PBS-1748](https://jira.il2.dso.mil/browse/PBS-1748) - Remove Scene Delegate Dependency


bmorris-revacomm (approved):
looks good 🌂

PR #565, opened 2021-08-24
PBS-1715 Fix flying org input

mcgarrys:
This PR fixes the strange behavior in the flying organization input. Previously, we were applying zeros while the user was typing, and using **.suffix** to enforce the character limit. This made it appear as if you could keep typing infinitely when there were only 4 characters. 

- Replaces **.suffix** with **.prefix** in **enforceFlyingOrgFormatting()**.
- Adds **textFieldDidBeginEditing()** and **textFieldDidEndEditing()** to the TabbedTextField's Coordinator.
- Adds the ability to pass closures into a TabbedTextField, and then into its Coordinator, to perform actions for editing state change events. These closures are optional.
- Factors out the logic applying zeros to the flying org input, and applies them when the textfield's editing session has ended. 
- Reorganize some of the functions in the String+Ext file.

[PBS-1715](https://jira.il2.dso.mil/browse/PBS-1715) - Fix Flying Org Input

- [x] Ensure the flying org textfield character limit behaves the same as the ICAO and SSN textfields (you are unable to input any new characters after inputting the 4th character).
- [x] Enter less than 4 characters in the flying org textfield and tap on another textfield, or close the keyboard. Ensure zeros are added in front of whatever you just inputted so that the length of the text is now 4.

### Notes :writing_hand: 
- I was debating what I should name the optional closures being passed down to the TabbedTextField's Coordinator. Settled with "didBeginEditing" and "didEndEditing" because it reads like the delegate function does at the call site. Let me know what you think.


jbethanco (approved):
Works great here.

PR #566, opened 2021-08-24
Fix/first name tab - (PBS-1739)

bmorris-revacomm:
Fix aircrew list textfield tabbing

refactor for each to correctly assign tab indices for textfields

[PBS-1739](https://jira.il2.dso.mil/browse/PBS-1739)

- [ ] Test that tabbing goes in correct order and continues to next row


jbethanco (approved):
This works. I thought adding a new row might break it, but it still worked. Cool.

PR #567, opened 2021-08-25
Change Initials field to tabbed text field.

jbethanco:
Change initials field to tabbed text field

[PBS-1756](https://jira.il2.dso.mil/browse/PBS-1756) 

- [x] Using the keyboard chevrons... can tab through to the initials
- [x] If the sortie has an aircraft commander with a first and last name, and the initials field is blank, when the initials field gets focus, it adds the initials automatically.


mcgarrys (approved):
Works great here 🔥. Out of curiosity, why are we adding a placeholder for this field?

jbethanco:
> Works great here 🔥. Out of curiosity, why are we adding a placeholder for this field?

I don't know.

PR #568, opened 2021-08-25
Fix/refactor initializer

jbethanco:
I had this laying around in a stash after removing the scene delegate/mission from the inits.
This moves some calculations out of the initializer and into computed vars. 
Really curious your guys thoughts on this. I like it because it reduces lines of code by about half. But that is about it. 

Removes a couple of initializers in favor or computed vars. 

None

- [ ] Do the touched PDFs still generate correctly?


bmorris-revacomm (approved):
SHIP IT

mcgarrys (approved):
I like it. The code reduction is nice, and it's still clear what's going on. Doesn't seem to affect anything functionality-wise. 🧙‍♂️

PR #582, opened 2021-08-30
Fix/Default Crew Lines

bmorris-revacomm:
New sorties need to add all crew lines from all previous sorties without any duplicates

- add func on mission for getting all unique crewlines
- write unit test for func

[PBS-1784](https://jira.il2.dso.mil/browse/PBS-1784)

- [x] Make sure new sorties add all crew lines from all previous sorties without any duplicates


jbethanco (commented):
Works good... one minor change...  I'm thinking we should check if > 1 AC, and if so, remove all ACs. I'm thinking none is better than two. 
image

jbethanco (approved):
Works.

PR #585, opened 2021-08-31
PBS-1753 Fix AircrewRow double initialization bug

mcgarrys:
This PR fixes the double initialization bug that was occurring with AircrewRow. The issue was that we were saving the sortie every time the **AircrewStaticListView** disappeared, which would make the views using the sortie reinitialize. 

- Removes the **DataController** save from **AircrewStaticListView** and adds it to **AircrewListView**, where we are actually doing the editing.

[PBS-1753](https://jira.il2.dso.mil/browse/PBS-1753)

- [x] Add a print statement to the **AircrewRow** initializer and ensure it gets called once for every CrewLine in the sortie.
- [x] Make a change to a CrewLine and make sure the data is saved properly after navigating away.


jbethanco (approved):
Works.

PR #587, opened 2021-09-01
Create index for crewLines

jbethanco:
When adding multiple rows to aircrew, typing in one field was sometimes duplicated in other fields on a different row. e.g. changing the first name on one row would update the first name on another row. Previous attempts at fixing this required global vars, singletons, changing view state while updating, resulting in purple errors, and general view lag when adding members. 

This approach seems to resolve all the things. 

all-the-things-hyperbole-and-a-half

Add transient property on crewLine to hold an index value.
Set crewLine index in the sortie.crewLines accessor.
Use crewLine index for tab order in a proper ForEach loop for AircrewRows.

[PBS-1802](https://jira.il2.dso.mil/browse/PBS-1802) Duplicate CrewMember text bug

- [x] Add multiple aircrew rows, type in the fields, make sure it isn't duplicated 

### Resources :books: 
https://developer.apple.com/wwdc21/10022

image


mcgarrys (approved):
Works great 💫

PR #588, opened 2021-09-01
Change to SwiftUI App Lifecycle

jbethanco:
Moving to SwiftUI App Lifecycle as a step in the direction of controlling the world as discussed. 
No "Current" yet. 
AppDelegate had to remain in support of disabling third party keyboards which is a NowSecure identified vulnerability. 
The app delegate is declared in the PBLoggingApp file with the UIApplicationDelegateAdaptor property wrapper.
It looks like an unused var, but apparently that is how it is done so the app knows what to use as the app delegate. 

Delete SceneDelegate
Add PBlogginApp App protocol file
Update app delegate for App lifecycle usage
Add app delegate to App file with property wrapper
Update UI test to pass in separate launch argument for passcode
Update shouldAllowExtensionPointIdentifier comment

[PBS-1393](https://jira.il2.dso.mil/browse/PBS-1393)

- [ ] Make sure the current 1.1.0 is installed. Add some missions and other data.  Switch to this branch and build. Make sure app starts and all previous data is still there. 
- [ ] This guts how the app starts and where/when objects are initialized. Please give the entire app a once over and ensure everything is operating the same as the previous build. Hopefully the only differences are a little speed boost. 

### NowSecure Score 🔒
100%


bmorris-revacomm (approved):
I tried adding several missions and sorties and filling out all the information for the sortie - everything was still there when I switched to the other branch - can't test upload or import cuz no IL4

PBLogging/Info.plist
```diff
@@ -31,19 +31,7 @@
...
-			<array>
-				<dict>
-					<key>UISceneConfigurationName</key>
-					<string>Default Configuration</string>
-					<key>UISceneDelegateClassName</key>
-					<string>$(PRODUCT_MODULE_NAME).SceneDelegate</string>
-				</dict>
-			</array>
-		</dict>
+		<true/>
```
mcgarrys, line 34: Is there a reason why UIApplicationSupportsMultipleScenes was flipped to true? Does it need to be set to true when a SceneDelegate class isn't present?

This is what Apple says about setting that key to true:

Screen Shot 2021-09-01 at 7 05 25 PM

mcgarrys (commented):
Seems to work well, although I left a clarification comment in the Info.plist file.  

I tested pretty heavily, filling in every field we have and adding more than one of everything. Switched to your branch and it was all there, PDF still generated. 🧠  🔥 🦄

PBLogging/Info.plist
jbethanco: Without it, had to delete the app when updating to the App Lifecycle branch. It may be something we can switch off after a few weeks once everyone has updated.  We only have the one scene, so we should be good.

PR #591, opened 2021-09-03
Make Login Process More Reliable

jbethanco:
Login page would sometimes hang after entering OTP. 
This cause that page to reload without user interaction.
When a user logged in or various server errors happened, the user was presented with a blank page. This adds a message to indicate what might be happening.

Add autoscroll to bottom of acceptance page
Add automatic reload for hung pages
Add message for server problems or non-IL4 access
Add extension on WKWebView to make setting and getting values from WKWebView easier. 

Import page stays up and blank after logging in
https://jira.il2.dso.mil/browse/PBS-1755
Make login process more reliable
https://jira.il2.dso.mil/browse/PBS-1790

- [x] If you don't have IL4 access, attempt login to the download events view and if you get a red message at the top... it worked.
- [ ] If you have IL4 access, login multiple times and see if the process is smoother than before.


mcgarrys (approved):
Works great on IL2, tupled and saw it works on IL4 🍻

PR #592, opened 2021-09-03
PBS-1813 Update 791 generation to print zeros

mcgarrys:
This PR updates the 791 PDF generation so that "0.0" is printed if there is no amount set for quantity issued pounds on an issue. Before, if quantityIssuedPounds was 0, it would show up blank on the PDF. 

- Adds **quantityIssuedPoundsString** to Issues+Accessors
- Implements this String in the 791 generation instead of the raw **quantityIssuedPounds** so that the "0.0" actually shows up (The **setText** method for Doubles has a guard let that will return early if the value is 0).

[PBS-1813](https://jira.il2.dso.mil/browse/PBS-1813) - On 791 if issue amount is zero print 0 instead of blank

- [ ] Navigate to any tanker mission, add an issue, and leave quantity issued as "0.0". Generate the PDFs and ensure that the "0.0" shows up under "Pounds" on the 791 Form.


jbethanco (approved):
Works.

PR #594, opened 2021-09-08
PBS-1819 Remove 0s from Combat/Combat Support on 781

mcgarrys:
This PR address the issue that Leah mentioned in our meeting today. On the 781, zeroes would appear under Combat/Combat Support Sortie count, when there should be nothing there. 

- Adds **shouldDisplayZero** parameter to **setText** for Int16s. Allows us to pass in a Int16, and if we want to show nothing when it is 0, we can just set **shouldDisplayZero** to false.
- Implements this on **flightConditionsCombatSorties**, and **flightConditionsCombatSupportSorties** when setting text in the PDF generator.

[PBS-1819](https://jira.il2.dso.mil/browse/PBS-1819) - Remove 0s from Combat/Combat Support on 781

- [ ] Navigate to any mission, and make sure Combat or Combat Support is _not_ set as the sortie type. Generate the 781s, and make sure nothing shows in the corresponding Combat and Combat Support sortie columns.
- [ ] Navigate to another mission, select Combat as the sortie type, and hit the gonkulate button. Generate the 781s, and make sure values appear in the corresponding Combat Time & Combat Sortie columns. 
- [ ] Repeat the previous step, this time selecting Combat Support as the sortie type. Generate the 781s, and make sure vales appear in the corresponding Combat Support Time & Combat Support Sortie columns.


jbethanco (approved):
Works. 🧠

PR #595, opened 2021-09-08
PBS-1820 Switch positions of "To" and "From" on sortie card

mcgarrys:
This PR addresses the feedback we got from Leah today about the positions of the takeoff and land ICAOs on the sortie card being confusing. 

- Switches "To" and "From" ICAOs on the sortie card.

[PBS-1820](https://jira.il2.dso.mil/browse/PBS-1820) - Switch positions of To and From on Sortie card

- [x] Ensure "To" appears above "From" on sortie cards.


jbethanco (approved):
Works... wondering if we should change on mission data in order to be consistent.

mcgarrys:
> Works... wondering if we should change on mission data in order to be consistent.

That sounds like a good idea.

PR #596, opened 2021-09-08
Fix Download PBS Event Bug

jbethanco:
User reported that they were unable to download any missions from prod even though they had multiple missions.

Make organizationID optional on PBSEvent to prevent decode failure that was causing the problem. 
Add alert for download event decode errors
Change Format slightly in PBSStructs

https://jira.il2.dso.mil/browse/PBS-1737 

- [x] Download missions for said user, using their pbsID. In the API request builder, insert the following line after line 79 :
components.path = "/puckboard-api/v1/pbl/crewMember/1ee6c8bd-0a50-41cc-80f3-571f5ae5dd35/events"

- [x] Ensure your missions download as per usual

If on IL2, let me know and I can tuple with you to demonstrate the change/effect


mcgarrys (approved):
Tupled and saw that it works 🪐

PR #599, opened 2021-09-14
Feature/feedback overlay

bmorris-revacomm:
Create native feedback form that overlays the original one. Use @Binding/javascript to set up communication between the two feedback forms' elements.

- add feedback form view
- inject javascript on collector web view
- add extension on String for removing prefix and suffix
- refactor content button

[PBS-1170](https://jira.il2.dso.mil/browse/PBS-1170)
[PBS-1293](https://jira.il2.dso.mil/browse/PBS-1293)
[PBS-1145](https://jira.il2.dso.mil/browse/PBS-1145)

- [ ] Check that you cannot submit unless everything is filled out (excluding images)
- [ ] Check that all data uploads correctly to Jira
- [ ] Check that a removed image does not upload to Jira
- [ ] Check that an image that was unchecked then rechecked is uploaded to Jira
- [ ] Check for any UI bugs

### UI Changes :eye: 
Screen Shot 2021-09-14 at 10 37 29 AM

### Resources :books: 
- Google
- my brother


jbethanco (commented):
Need to disable autocomplete on the textfields so that we don't lose our nowsecure 100. 
What do you think about adding last username and email to the defaults? Not sure what that would do to nowsecure. 
I was able to lock up the UI by clicking stuff while the photo was uploading. I'm wondering if we can use the image call back with "in progress" message to lock the UI while the upload completes and show the user a message showing it is in progress." Then unlock once we get the normal callback.

bmorris-revacomm:
> Need to disable autocomplete on the textfields so that we don't lose our nowsecure 100.
> What do you think about adding last username and email to the defaults? Not sure what that would do to nowsecure.
> I was able to lock up the UI by clicking stuff while the photo was uploading. I'm wondering if we can use the image call back with "in progress" message to lock the UI while the upload completes and show the user a message showing it is in progress." Then unlock once we get the normal callback.

Not sure about user defaults.... I'm thinking maybe not doing that, just to make sure we always have the correct person and to avoid now secure issues. Plus, I don't think people give feedback often enough for it to be a hassle to fill it out every time. But if you disagree, we can put it in!

jbethanco (approved):
Works great. 🤳

PR #602, opened 2021-09-16
Remove Sheets from Menu Item Bars

jbethanco:
Sheets attached to buttons inside Nav View menu item bars had an odd effect of losing track of the apps presentation mode. 

Removes the sheets from the menu items and adds them to views outside the menu items. 
Adds an app state object we can use for other things, but should probably replace with the Control the World "Current" soon. 

https://jira.il2.dso.mil/browse/PBS-1893

- [ ] Bring up a sheet, close out of the app, wait a few seconds, return to the app, wait a few seconds, close the sheet. Repeat a few times. If that works, this PR works. If the sheets fail to close, this PR fails.


bmorris-revacomm (approved):
works for me 🧑‍🎄

PR #605, opened 2021-09-20
Handle 403s from PBS uploads

jbethanco:
PBS is handling PBL permissions to upload on their end. If a unit doesn't have access to upload, they receive a 403 with a specific message. This handles that error and displays a message indicating such.

Catch and handle 403 PBS status codes and display permission message
Add PBSErrorMessage struct
Add hasAccess boolean to the upload view model
Move Result Message Text back into view from var so it updates

https://jira.il2.dso.mil/browse/PBS-1834
 
### UI Changes :eye: 
Simulator Screen Recording - _iPadPro(9 7) - 2021-09-20 at 09 23 07


PBLogging/Network/UploadMissionViewModel.swift
```diff
@@ -63,6 +64,11 @@ class UploadMissionViewModel: ObservableObject {
                     }
             case .failure(let error):
                 switch error {
+                case .noPermissions:
+                    DispatchQueue.main.async {
+                        self.hasAccess = false
+                        self.message = "Puckboard Logging is not yet enabled for your organization. To join the waitlist, please reach out to the Puckboard team."
```
mcgarrys, line 70: Was going to suggest adding the error messages to the Strings file, but then I had the thought of adding a "message" var to the NetworkClientError. Would look something like this.

Screen Shot 2021-09-20 at 5 16 37 PM

And then at the call site, we can have to assign the error message to the published var just once, rather than repeating that in every case. The switch statement would just be used to perform any leftover logic.

Screen Shot 2021-09-20 at 5 19 55 PM

Thoughts? I may be forgetting/missing something.

mcgarrys (approved):
Looks great. Left a few comments, including a suggestion that could make accessing error messages at call sites cleaner. Curious what your thoughts are.

PBLogging/Network/UploadMissionViewModel.swift
mcgarrys: And we could still add the Strings to the Strings constants file, and access them in the enum.

PBLogging/Network/UploadMissionView.swift
```diff
@@ -46,7 +45,20 @@ struct UploadMissionView: View {
             
             Group {
                 uploadButton
-                uploadSuccessMessage
+                    .disabled(!model.hasAccess)
+                Text(model.message)
+                    .font(.pblBold(size: defaultFontSize))
+                    .foregroundColor(model.hasAccess ? .pblForegroundPrimary : .red)
```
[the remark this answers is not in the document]
jbethanco: done

PBLogging/Network/UploadMissionView.swift
```diff
@@ -66,7 +78,7 @@ struct UploadMissionView: View {
         .background(Color.pblBackgroundDefault)
         .alert(isPresented: $showAutoInitialAlert) {
             Alert( title: Text("Apply Initials?"),
-                   message: Text("Would you like to apply initials to all sorties based on each sortie's aircraft commander's first and last name? "),
+                   message: Text("Would you like to apply initials to all sorties based on each sortie's aircraft commander's first and last name?"),
```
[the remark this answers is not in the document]
jbethanco: Done

PBLogging/Network/UploadMissionViewModel.swift
jbethanco: I want to relook at the whole networking topic as a whole once the download missions thing is complete. 
Also want to look at maybe moving away from the strings constants file. It sounded like a good idea at the time, but now it seems it just moves things a step away from where you need it when you want to edit.

PR #614, opened 2021-09-28
Fix/button bug

bmorris-revacomm:
- add button style so entire row is not tappable (go figure)

[PBS-1918](https://jira.il2.dso.mil/browse/PBS-1918)

- [ ] Make sure tapping generate forms button is the only thing that triggers the button - not tapping anywhere on the row


jbethanco (approved):
Wow.

PR #639, opened 2022-01-27
WIP - Log in and Sync with Puckboard, Network Code Refactor

Adam-Roney581:
This is a work-in-progress pr because right now, the only thing that "syncs" are tail numbers. Should I comment out some stuff and merge? Are we ready for the Log in with Puckboard button? 

You can now log in with your P1 account and it shouldn't ask you to log in again. You can still log in the old way and the log in prompt will work the same as it did before. I also added a "Sync with Puckboard" button under the Info Button. Right now, this only pulls tail numbers, but you should be able to easily add whatever information we deem necessary. 

I also re-factored some of the network code. I moved some stuff into the network client to remove duplicate code. 

[PBS-2101](https://jira.il2.dso.mil/browse/PBS-2101)
[PBS-2099](https://jira.il2.dso.mil/browse/PBS-2099)

- [x] Log in using the "Log in with Puckboard" button
- [x] Test Import / Upload and make sure it doesn't ask you to log in again and. it pulls the correct information

- [x] Log in regularly (Or skip on debug)
- [x] Test Import / Upload / Sync with Puckboard all require log in before showing the content

- [x] Log in regular or Log in with Puckboard
- [x] Tap on the info icon and select "Sync with Puckboard"
- [x] Verify that it shows you the log in page (if necessary)
- [x] Review UI for the sync load
- [x] Verify tail numbers were successfully pulled
- [x] Kill the app after logging in with p1, verify on next launch it logs you in
### Notes :writing_hand: 

Eventually, this login and sync should pull training accomplishments as well. (Or other information that might be relevant)

Right now, it pulls tail numbers anytime you log into P1. 

### UI Changes :eye: 
Added "Log in with Puckboard" button
Added a new "Sync with Puckboard" option under the Info Button
Added Sync view to show progress when syncing with Puckboard


PBLogging/Views/UtilityViews/SyncView.swift
```diff
@@ -0,0 +1,72 @@
...
+        .cornerRadius(10.0)
+        .onAppear {
+            startSync()
+        }
+        .showOnlyIfLoggedInToPBS()
+    }
+    
+    // TODO: Also pull training accomplishmnets. Update progress and message accordingly
+    // Using delays to simulate progress of pulling info from puckboard
+    func startSync() {
```
cbrech, line 37: Need to add detection/support for no and dropped internet connections

cbrech (changes requested):
Overall love it.

Seems to work well most of the time. Does lock up screen when you try and sync with no internet connection.

Maybe a cancel button too if it fails midway? 

If the intent is to lockout the whole screen, perhaps make the popover blur everything behind it?

Also need to make sure it presents a good error message when the user logging in is part of an org that does not yet have PBL access.

How does it pick which organization a user logs into?

PBLogging/Views/UtilityViews/SyncView.swift
Adam-Roney581: I have the button in the info popover to be disabled when there's no internet connection. The sync view is also now in a closable sheet so it's easily closable if something goes wrong.

Adam-Roney581:
Added the sync view into a close-able sheet so it doesn't lock up the screen. Also added an organization selector to the sync view for when we're going to eventually sync training accomplishments. 

Added a check for the user's primary organization when logging in with puckboard and show's an alert if they don't have access yet.

I made the organization selector the it's own view so that it could be re-used and it should work the same way was before. 

Let me know if anything else needs to be changed.

cbrech:
Loving this. Noticed a bug when you switch from Staging to Prod or vice-versa on the Sync with Puckboard view. Looks like it needs detection for when the session with Puckboard is expired.

Screen Shot 2022-02-07 at 08 24 00

cbrech:
@Adam-Roney581 also looks like the app is now requiring a login on each launch. We only want them to have to login once because a) logins are annoying, and b) they need to be able to open the app in the air without internet.

Adam-Roney581:
Fixed the issue for sync logging you in (Forgot to add back the modifier that shows log in). Added a username when a user logs into p1 so that it keeps them logged in. Updated the design of the organization selector.

cbrech (approved):
Love it!

cbrech:
@Adam-Roney581 I think you can resolve the merge conflicts and then this should be merged. Lots of goodness in here that doesn't need to wait for more than tail numbers to be included.

PR #647, opened 2022-02-14
Duplicate Crew Members on 781 and Cumulative Crew Members Sortie Change

Adam-Roney581:
Do we want to only copy crew members from the last sortie? Right now it's cumulative but it's causing problems for people with large rotating crews with several sorties on one mission. 
 
Instead of looking at the crew members core data UUID, it is looking at an array of their first name, last name, and last social when comparing duplicates. This fixes the issue of duplicate entries of the same user on 781 generation. 

Pulls the crew from the most recent sortie instead of being cumulative of all sorties. This fixes the issue for missions with large, rotating crew members when creating several sorties. 

[PBS-2147](https://jira.il2.dso.mil/browse/PBS-2147)
[PBS-2148](https://jira.il2.dso.mil/browse/PBS-2148)

- [ ] Create a sortie with a crew member
- [ ] Create another sortie, delete the existing crew member and then add them back
- [ ] Generate 781 and you should see only one entry for that individual 

- [ ] Create a sortie with some crew members
- [ ] Create a new sortie and add/delete/edit crew members
- [ ] Create a new sortie and it should only contain the crew members from the previous (most recent) sortie


cbrech (changes requested):
Looks like we need to add/update the unit tests for this.

Adam-Roney581:
Should be good to go now.

cbrech (approved):

PR #659, opened 2022-03-24
Add Boom Operator Support

Adam-Roney581:
This adds boom operator support when the mission is a tanker mission (MDS starts with "K"). 
Adds boom operator support for time splitting and training accomplishments
If the mission is a tanker mission, loadmasters are added to the "Other" category
If the mission isn't a tanker mission, boom operators are added to the "Other" category
Time splitting for booms works the same way as loadmasters without the loadmaster stepper
[PBS-2175](https://jira.il2.dso.mil/browse/PBS-2175)

- [ ] Add boom operator to crew (IB, EB, MB)
- [ ] Check flight time view and verify that the booms are in the other category
- [ ] Go back and change the mds of the mission to KC-10 (Or K whatever)
- [ ] Check the flight time view and verify that the booms are in their own category and any loadmasters are in the other category
- [ ] Split some time and verify it works with the pdf generation 

### Notes :writing_hand: 
There are no boom training accomplishments yet, so it's an empty page as of right now. 
### UI Changes :eye: 
Added support for Boom Operators in the segmented controls for flight time and training accomplishments
Edited the training accomplishment page to better handle no training accomplishments


Adam-Roney581:
This is going to require some changes to add a "Required Boom Operators" stepper like the loadmasters page has. This will probably require a core data update

cbrech:
@Adam-Roney581 why base boom support on the K vs having people with boom duty codes (B)?

cbrech:
Other than two-stepper, looks great.

Adam-Roney581:
@cbrech "why base boom support on the K vs having people with boom duty codes (B)?" 

I think we need to know if it's a tanker sortie to know whether or not to show the Boom Operator choice, or to put them in the "Other" category. Same with loadmasters. Do you think it should always show booms as an option if they are added regardless if it's a tanker sortie or not?

cbrech:
> 

I think so, prob not booms but will have some other crew positions (like flight attendants, etc.) that can be on multiple airplanes so having one set of logic seems simpler.

Adam-Roney581:
Updated to show booms if they are added regardless if it is a tanker sortie or not. Added stepper for booms for the "numBoomsRequired" attribute that was added to core data. This will eventually require an api update so that we can push this new attribute. 

**Please test core data migration**

Adam-Roney581:
Sortie Data Model change on the web side

https://jira.il2.dso.mil/browse/PBS-2357

thardie2 (approved):

PR #667, opened 2022-04-06
Feature/loading indicator event download

xijiewang4:
added loading indicator for event download
nothing but shows circle spinning before mission are finished pulling 
[PBS-2294](https://jira.il2.dso.mil/browse/PBS-2294)

- [ ] Testing Item

### UI Changes :eye: 
spinning circle
### Resources :books: 
https://jira.il2.dso.mil/browse/PBS-2262


cbrech (changes requested):
Gets stuck on loading screen. See MM video for details

PBLogging/Network/GetUploadedMissionSheet.swift
```diff
@@ -37,6 +37,8 @@ struct SelectUploadedMissionView: View {
     @StateObject var model: GetUploadedMissionViewModel
     @State var isShowingAlert = false
     @State var selectedOrganization: String
+    @State var uploading = true
```
Adam-Roney581, line 40: Shouldn't this variable name be downloading since it's on a download view? I was a little confused at first

xijiewang4: good point, will make the change

Adam-Roney581:
It get's stuck in a loop of the "no permission" error when it hits it one time. Also, I think we may need another message for when there aren't any missions. For example: On prod, Tron Demo Squadron 2 doesn't have any missions to download, but it gives me the no permissions error. I have permission, there's just no missions in that squadron.

cbrech (changes requested):
Works great on the second try but fails on first attempt.

Steps to reproduce:
* Build app
* Click "Previously Uploaded Post Mission Data"
* Login
* Spinner keeps spinning and bouncing - never loads
* Exit screen 
* Click "Previously Uploaded Post Mission Data"
* Works great

cbrech (approved):

